PR #1, opened 2023-06-08
Basics

SimonHeybrock:
Very basic, this is for warmup.


jl-wynen:
Not sure what I should review here. Looks like the basic stuff we discussed.

SimonHeybrock:
@jl-wynen Have another look, I added some `dask` stuff.

src/sciline/container.py
```diff
@@ -0,0 +1,94 @@
...
+            raise UnsatisfiedRequirement(e) from e
+        return task if self._lazy else task.compute()
+
+
+def _delayed(func: Callable) -> Callable:
+    """
+    Decorator to make a function return a delayed object.
+
+    In contrast to dask.delayed, this uses functools.wraps, to preserve the
+    type hints, which is a prerequisite for injector to work.
```
jl-wynen, line 40: Do you know why dask doesn't use `wraps`? They might have a good reason.

src/sciline/container.py
```diff
@@ -0,0 +1,94 @@
...
+    #
+    # 1. Intermediate results are used more than once.
+    # 2. Intermediate results are large, so we generally do not want to keep them
+    #    in memory longer than necessary.
+    #
+    # To address these problems, we can internally build a graph of tasks, instead of
+    # directly creating dependencies between functions. Currently we use Dask for this.
+    # The Container instance will automatically compute the task, unless it is marked
+    # as lazy. We therefore use singleton-scope (to ensure Dask will recognize the
+    # task as the same object) and also wrap the function in dask.delayed.
```
jl-wynen, line 66: This comment should rather be part of a design document instead of hidden away in the source code.

src/sciline/container.py
```diff
@@ -0,0 +1,94 @@
...
+    scope = injector.singleton
```
jl-wynen, line 67: Does this ensure that the dask graph correctly identifies nodes with each other?

jl-wynen, line 67: Do you have to guarantee that all dependencies of this function are also singletons?

tests/container_test.py
```diff
@@ -0,0 +1,65 @@
...
+    assert hasattr(task, 'compute')
+    assert task.compute() == 1.5
+
+
+def test_lazy_with_multiple_outputs_computes_intermediates_once():
+    ncall = 0
+
+    def provide_int() -> int:
+        nonlocal ncall
+        ncall += 1
```
jl-wynen, line 58: Is this atomic? Might be worth setting a serial scheduler for dask.

src/sciline/container.py
SimonHeybrock: I suppose because it would be lie? The types are actually different.

src/sciline/container.py
SimonHeybrock: I can add it.

src/sciline/container.py
SimonHeybrock: Yes, that is what I have observed. I think it is related to https://docs.dask.org/en/stable/delayed-best-practices.html#avoid-repeatedly-putting-large-inputs-into-delayed-calls, even though I would have expected that I would still be able to tell that objects are the same after hashing.

SimonHeybrock: You mean, what if someone manually inserts other non-singleton providers into the container? It will still work, but you may get duplicate computation.

tests/container_test.py
SimonHeybrock: Good point. Actually `injector` should be taking care of this, it is using a lock.

src/sciline/container.py
jl-wynen: Do why does this code?

src/sciline/container.py
jl-wynen: What types did you test this with? Scipp containers don't support hashing and dask basically falls back to calling `id`.

jl-wynen: So if a singleton depends on a non-singleton, it would recompute the dependency? What about the singleton itself? Does that get recomputed?

src/sciline/container.py
SimonHeybrock: See also the comment in `Container.get`: `dask.Delayed` is not a generic, so we cannot use `dask.Delayed[T]`. This implies that `injector` could not do its job.

src/sciline/container.py
SimonHeybrock: `int` and `float`.

SimonHeybrock: > So if a singleton depends on a non-singleton, it would recompute the dependency?

No, why?

> What about the singleton itself? Does that get recomputed?

Singletons are computed only once.

jl-wynen: This is my point. The dependency of a singleton is essentially also a singleton. Except when another non-singleton provider depends on it.

This is fine as long as it always returns the same result. I don't know if there is a way to have providers that return a different value every time.
But at least with delayed, there are subtleties. Given
```py
def a() -> A: ...

@singleton
def b(x: A) -> B: ...

def c(x: A, y: B) -> C: ...
```
calling `get(C)` will result in two separate `Delayed[A]` objects.

SimonHeybrock: Yes, because `A` is not a singleton. Not sure what your point is?

jl-wynen: I want to clarify the interactions between `inejctor`, our wrapper, and dask. There may be a lot of edge cases that we need to get right.
So when we automatically set providers as singletons, we should likely prevent any other kind of binding by the user.

SimonHeybrock: I don't think I can answer this at this point, without developing more functionality.

jl-wynen (approved):
Just go ahead. We'll probably come back and make changes anyway.

PR #4, opened 2023-07-10
Parameterized domain types --- a mechanism for sub-workflow templates

SimonHeybrock:
Initial attempts at making use of child-containers (child injectors) for support sub-workflows proved challenging: I ran into both technical issues (`injector` did not behave as I had assumed when working with child injectors) as well as UX ones (the client code was complex to see through, even for me who came up with it).

Therefore, I moved to a more explicit and "pedestrian" approach. I should be easier to see through.

The main challenge here was to find a way of defining "generic" domain types. We want to use `NewType` to, e.g., turn a `scipp.DataArray` into something specific, but would like to use a syntax such as `IofQ[Sample]` to refer to the `IofQ` of the "Sample" run. `parametrized_domain_type` is the current result of this.


src/sciline/domain.py
```diff
@@ -0,0 +1,33 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from typing import Dict, NewType
+
+
+def parametrized_domain_type(name: str, base: type) -> type:
+    """
+    Return a type-factory for parametrized domain types.
+
+    The types return by the factory are created using typing.NewType. The returned
```
jl-wynen, line 10:
```suggestion
    The types returned by the factory are created using typing.NewType. The returned
```

src/sciline/domain.py
```diff
@@ -0,0 +1,33 @@
...
+    ----------
+    name:
+        The name of the type. This is used as a prefix for the names of the types
+        returned by the factory.
+    base:
+        The base type of the types returned by the factory.
+    """
+
+    class Factory:
+        _subtypes: Dict[str, type] = {}
```
jl-wynen, line 24: So `parametrized_domain_type` must only be called once per `name`? That is, does using `Monitor = parametrized_domain_type("Monitor", DataArray)` in two different files lead to distinct types?

tests/complex_workflow_test.py
```diff
@@ -0,0 +1,105 @@
...
+IncidentMonitor = sl.parametrized_domain_type('IncidentMonitor', float)
+TransmissionMonitor = sl.parametrized_domain_type('TransmissionMonitor', float)
+TransmissionFraction = sl.parametrized_domain_type('TransmissionFraction', float)
+IofQ = sl.parametrized_domain_type('IofQ', np.ndarray)
+BackgroundSubtractedIofQ = NewType('BackgroundSubtractedIofQ', np.ndarray)
+
+
+def reduction_factory(tp: type) -> List[Callable]:
+    def incident_monitor(x: Raw[tp]) -> IncidentMonitor[tp]:
+        return IncidentMonitor[tp](x.monitor1)
```
jl-wynen, line 38: It would be nice if we could avoid repeating the return type. The injector should be able to convert the return value to the correct type. (Would likely need modifications to injector package.) But would this mess up type hinting?

src/sciline/domain.py
SimonHeybrock: As far as I can tell this is the same for `typing.NewType`?

tests/complex_workflow_test.py
SimonHeybrock: At least for run-time, you can just omit it. I don't know if the type-checker needs it.

src/sciline/domain.py
jl-wynen: I think so. I just wanted to make sure. This looks like the right thing to do.

tests/complex_workflow_test.py
```diff
@@ -0,0 +1,105 @@
...
+Masked = sl.parametrized_domain_type('Masked', np.ndarray)
+IncidentMonitor = sl.parametrized_domain_type('IncidentMonitor', float)
+TransmissionMonitor = sl.parametrized_domain_type('TransmissionMonitor', float)
+TransmissionFraction = sl.parametrized_domain_type('TransmissionFraction', float)
+IofQ = sl.parametrized_domain_type('IofQ', np.ndarray)
+BackgroundSubtractedIofQ = NewType('BackgroundSubtractedIofQ', np.ndarray)
+
+
+def reduction_factory(tp: type) -> List[Callable]:
+    def incident_monitor(x: Raw[tp]) -> IncidentMonitor[tp]:
```
SimonHeybrock, line 37: `mypy` does not like this: `tp` is a variable (holding a type) and cannot be used for type-checking.

tests/complex_workflow_test.py
jl-wynen: Tried it and neither mypy nor pyre like it when I just return the base type. So I guess we need to explicitly wrap the result.

tests/complex_workflow_test.py
jl-wynen: You could try 
```python
def reduction_factory(tp: type) -> List[Callable]:
    T = TypeVar("T", bound=tp)

    def incident_monitor(x: Raw[T]) -> IncidentMonitor[T]:
```

But this probably doesn't work either.

SimonHeybrock:
I think https://github.com/python/mypy/issues/3331 might help to resolve these problems, but it has seen little progress over time.

SimonHeybrock:
Closing, in favor of a slightly different approach.

PR #6, opened 2023-07-11
Support injecting via generics

SimonHeybrock:
This replaces #4.

Details in `Container` are a proof of concepts and likely still wrong in some cases. Will refactor and add tests after initial discussion. `complex_workflow_test.py` provides an example of how actual client code might look. Note the slightly verbose creation of parametrized/scoped domain types, but the simple/trivial way of using functions with generics in their type hints with `Container`.


src/sciline/container.py
```diff
@@ -48,47 +37,90 @@ def wrapper(*args, **kwargs):
...
+    def call(self, func: Callable[..., Return], bound: Optional[Any] = None) -> Return:
+        tps = get_type_hints(func)
+        del tps['return']
+        args: Dict[str, Any] = {}
+        for name, tp in tps.items():
+            if isinstance(tp, TypeVar):
+                tp = tp if bound is None else bound
+            elif (origin := get_origin(tp)) is not None:
+                if isinstance(get_args(tp)[0], TypeVar):
+                    tp = origin[bound]
```
YooSunYoung, line 82: Shouldn't it also check if the number of ``get_args`` is exactly 1 here as well?

SimonHeybrock: I had assumed that the check in `insert` takes care of that, but there is indeed some inconsistency here that I need to fix.

YooSunYoung: If all domain specific generic types will be sub class of `sl.Scope`,
we can use `isinstance(origin, sl.Scope)` instead of checking the length of `get_args` perhaps...?

But it may not work either since a subclass can have more type arguments than the parent class unless `sl.Scope` forces sub classes not to have more than the same number of type arguments via `__init_subclass__`....

SimonHeybrock: I am in the process of managing to make an improvement to the mechanism, by handling specific `TypeVar` instances as distinct when binding params of generics. This will ensure that functions provided by the user are not unintentionally bound together, and will furthermore allow for handling generics with multiple params (or multiple generics with distinct params). Will provide an update later.

SimonHeybrock:
I pushed a major update, the mechanism is now more or less at a point where I wanted to get. This is ready for a more detailed review.

src/sciline/container.py
```diff
@@ -86,19 +86,19 @@ def insert(self, provider: Provider) -> None:
...
-        return dask.delayed(func)(**args)  # type: ignore
+        args = {
+            name: self._get(self._bind_free_typevars(tp, bound=bound))
+            for name, tp in tps.items()
+        }
+        return dask.delayed(func)(**args)
+
+    def _bind_free_typevars(self, tp: type, bound: Dict[TypeVar, Any]) -> type:
+        if isinstance(tp, TypeVar):
+            return bound[tp]
```
YooSunYoung: Shouldn't it also be `bound.get(tp, tp)` in case it's not bound to anything, or is that not possible at all?

src/sciline/container.py
```diff
@@ -14,81 +25,128 @@ class UnsatisfiedRequirement(Exception):
...
+    return True
 
-    @wraps(func)
-    def wrapper(*args, **kwargs):
-        return dask.delayed(func)(*args, **kwargs)
 
-    return wrapper
+def _bind_free_typevars(tp: type, bound: Dict[TypeVar, type]) -> type:
+    if isinstance(tp, TypeVar):
+        return bound[tp]
```
YooSunYoung, line 50: Shouldn't it also be bound.get(tp, tp) in case it's not bound to anything, or is that not possible at all?

SimonHeybrock: I want it to fail in this case, so `__getitem__` works (to get `KeyError`). I could use what you suggest, to get an `UnsatisfiedRequirement` elsewhere. Not sure which is better? Or maybe we should have another specific exception like `UnboundTypeVar`?

YooSunYoung: I think raising `UnboundTypeVar` makes more sense because that's what you'll get from static type checker also.

src/sciline/container.py
```diff
@@ -14,81 +25,140 @@ class UnsatisfiedRequirement(Exception):
...
+            raise ValueError(f'Provider {provider} lacks type-hint for return value')
+        # isinstance does not work here and types.NoneType available only in 3.10+
+        if key == type(None):  # noqa: E721
+            raise ValueError(f'Provider {provider} returning `None` is not allowed')
+        if (origin := get_origin(key)) is not None:
+            subproviders = self._subproviders.setdefault(origin, {})
+            args = get_args(key)
+            if args in subproviders:
+                raise ValueError(f'Provider for {key} already exists')
+            subproviders[args] = provider
```
YooSunYoung, line 96: Should we forbid this kind of nested generic arguments...?

```
VT = TypeVar("VT")
Bike = NewType("Bike", object)
Car = NewType("Car", object)

class Time(sl.Scope[VT], float): ...

def provide_bike_time() -> Time[Bike]:
  return Time(1)
def provide_car_time() -> Time[Car]:
  return Time(0.1)

def provide_list_of_time(bike_time: Time[VT]) -> List[Time[VT]]
  return [bike_time]
```

It should be 
```
TimeT = TypeVar("TimeT")
def provide_list_of_time(bike_time: TimeT) -> List[TimeT]
  return [bike_time]
```
And then we can compute nested generic type like; `List[Time[Car]]` or `List[Time[Bike]]`.

SimonHeybrock:
> If there is a provider that loads file and if the file is replaced, we need to rebuild the dask delayed again. In that case, should we just rebuild the `Container` ...? Or do you want to have an interface that clears the `Container._cache`??

I don't think we need to do anything? `Container._cache` is not caching results, it is caching tasks. If you run the pipeline again, it should load the file again.

YooSunYoung:
> 

Now I remembered that once the `delayed.compute()` is done, it will no long using the previous intermediate results...
Then it should be fine indeed.
It will still use the same function even if we replace any existing provider,
but it's not allowed anyways so it is not an issue either I guess...?

PR #10, opened 2023-07-13
Support instances as "providers"

SimonHeybrock:


YooSunYoung:
I think retrieving `type` from an instance can cause unexpected errors.
How about having a separate interface like `insert_instance(providing_tp: type, instance: object): ...` and then returns the `instance` when it's requested.

Or just have a wrapper function that returns a function with the desired type?

Sth like...
```
def instance_provider(return_tp: type, instance: object):
  def _call():
    return instance
  _call.__annotations__['return'] = return_tp
  return _call

zero_provider = instance_provider(int, 0)
```

SimonHeybrock:
Closing in favor of different approach.

PR #11, opened 2023-07-14
Support instance "providers"

SimonHeybrock:
Replaces previous flawed approach from #10.

### DIFF
--- a/tests/pipeline_test.py
+++ b/tests/pipeline_test.py
@@ -388,3 +388,29 @@ def test_TypeVar_params_track_to_multiple_sources() -> None:
         [provide_int, provide_float, provide_A, provide_B, provide_C]
     )
     assert pipeline.compute(C[int, float]) == C[int, float](1, 2.0)
+
+
+def test_instance_provider() -> None:
+    Result = NewType('Result', float)
+
+    def f(x: int, y: float) -> Result:
+        return Result(x / y)
+
+    pl = sl.Pipeline([3, 2.0, f])
+    assert pl.compute(int) == 3
+    assert pl.compute(float) == 2.0
+    assert pl.compute(Result) == 1.5
+
+
+def test_instances_of_generics_cannot_be_distinguished_and_raise() -> None:
+    T = TypeVar('T')
+
+    @dataclass
+    class A(Generic[T]):
+        value: T
+
+    instances = [A[int](3), A[float](2.0)]
+
+    with pytest.raises(ValueError):
+        # Both instances have type A
+        sl.Pipeline(instances)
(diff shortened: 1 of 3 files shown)


tests/pipeline_test.py
```diff
@@ -388,3 +388,98 @@ def provide_C(x: A[T1], y: B[T2]) -> C[T1, T2]:
...
+    assert pl.compute(float) == 2.0
+    assert pl.compute(Result) == 1.5
+
+
+def test_provider_NewType_instance() -> None:
+    A = NewType('A', int)
+    pl = sl.Pipeline([])
+    pl[A] = A(3)
+    assert pl.compute(A) == 3
+
```
YooSunYoung, line 412: Maybe this kind of test case can be useful...?

```suggestion

def test_provider_Newtype_instance_same_supertype() -> None:
    pl = sl.Pipeline()
    Str = NewType('Str', str)
    IntStr = NewType('IntStr', str)
    pl[IntStr] = Str('0')
    assert pl.compute(IntStr) == '0'

```

SimonHeybrock: I suppose this is just a "limitation" of how `NewType` works. I think the type checker should be able to catch this?

SimonHeybrock: `mypy` does not, apparently. :(

YooSunYoung: This one is not caught by `mypy` or `pyre` either.
`TypeVar` seems quite loose if it's not used in `Generic` ...
If your intention is not allowing this at all, maybe adding it to the document should be sufficient...?

```
from typing import NewType, Dict, Type, TypeVar, Any
A = NewType('A', str)
B = NewType('B', str)

T = TypeVar("T")

class C:
    def __init__(self) -> None:
        self.type_dict: Dict[Any, Any] = dict()
    
    def __setitem__(self, tp: Type[T], _object: T) -> None:
        self.type_dict[tp] = _object

c = C()
c[A] = B('str')
c[int] = '0'
```

SimonHeybrock: > If your intention is not allowing this at all, maybe adding it to the document should be sufficient...?

Not allowing what?

YooSunYoung: To do this, setting an item of type `Str` as a provider of `IntStr` in this example.
```
 pl = sl.Pipeline()
Str = NewType('Str', str)
IntStr = NewType('IntStr', str)
pl[IntStr] = Str('0')
assert pl.compute(IntStr) == '0'
```

YooSunYoung (approved):

tests/pipeline_test.py
SimonHeybrock: Well, I don't see how I could prevent that, except by disallowing `NewType` as key?

PR #12, opened 2023-07-17
Refactor to split off dask and generalize graph builder

SimonHeybrock:
This refactors `Pipeline`, adding `build()` which returns a generic graph (currently as a `dict`). This can be turned into a task graph that works with `dask`, or can be used to make `graphviz.Digraph`. The internal use of `dask.delayed` is removed completely.

I think I am way more happy with this split. This opens the doors for, e.g., dropping the `dask` requirement completely, or implement a custom `dask` Collection on top of this.

I am not fully happy with the interface yet. I think what we have here now can be suitable as a low-level interface that we can build on top of. I am considering, among other things, splitting the `graph` into a class that providers some helpers such as `compute()`, or facilities to build a dask graph. Not sure exactly where this is going, but it may be useful to review now before such changes.


src/sciline/pipeline.py
```diff
@@ -187,31 +173,40 @@ def _get_provider(
...
-        return self._cache.setdefault(tp, delayed(**args))
+        tps = get_type_hints(provider)
+        args = {
+            name: _bind_free_typevars(t, bound=bound)
+            for name, t in tps.items()
+            if name != 'return'
+        }
+        graph: Graph = {tp: (provider, bound, args)}
+        for arg in args.values():
+            graph.update(self.build(arg))
```
jl-wynen, line 199: Potential infinite loop here (or up to max stack depth). @YooSunYoung has a solution based on `graphlib`.

src/sciline/graph.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def _format_graph(graph: Graph) -> Dict[str, Tuple[List[str], str]]:
+    return {
+        _format_provider(provider, bound=bound): (
+            [_format_type(a) for a in args.values()],
+            _format_type(ret),
+        )
+        for ret, (provider, bound, args) in graph.items()
+    }
```
jl-wynen, line 47: Not hugely relevant since graphs are small, but could yield tuples here instead of building a temporary dict.

src/sciline/graph.py
```diff
@@ -0,0 +1,70 @@
...
+def _format_provider(provider: Callable[..., Any], bound: Dict[TypeVar, type]) -> str:
+    s = provider.__qualname__
+    if bound:
+        s += f'[{",".join([_format_type(b) for b in bound.values()])}]'
+    return s
+
+
+def _format_type(tp: type | NewType) -> str:
+    """
+    Helper for Pipeline.get_graph.
```
jl-wynen, line 59: Out of date

src/sciline/graph.py
```diff
@@ -0,0 +1,70 @@
...
+    get_args,
+    get_origin,
+)
+
+from graphviz import Digraph
+
+from .pipeline import Graph
+
+
+def to_graphviz(graph: Graph) -> Digraph:
```
nvaytet, line 20: I think it would be a good idea to also add a test for this.

In addition, do we want to make it explicit that it is using `graphviz` (as it is currently), or is that an implementation detail that the user should not care about? Meaning we would call it something like `show_graph`, as we have done for the `transform_coords` and `plopp`.

jl-wynen (commented):
Good idea!
I agree that there should be a class for `Graph`. In particular since it currently is represented in terms of a relatively complicated tuple. I'd even have a separate graph class that only has the encoding and graph primitives in addition to a task graph class that has the higher level ops like `compute`.

tests/pipeline_test.py
```diff
@@ -483,3 +481,42 @@ def func(x: int, y: float) -> str:
 
     pl = sl.Pipeline(providers=[func], params={int: 1, float: 2.0})
     assert pl.compute(str) == "1 2.0"
+
+
+def test_build_graph_to_dsk() -> None:
```
nvaytet, line 486:
```suggestion
def test_build_graph_to_dask() -> None:
```

tests/pipeline_test.py
```diff
@@ -483,3 +481,42 @@ def func(x: int, y: float) -> str:
...
+    pipeline = sl.Pipeline([int_to_float, provide_int])
+    graph = pipeline.build(float)
+    dsk = {tp: (provider, *args.values()) for tp, (provider, _, args) in graph.items()}
+
+    assert dask.get(dsk, [float, int]) == (1.5, 3)
+
+    assert ncall == 1
+
+
+def test_build_graph_to_dsk_combine() -> None:
```
nvaytet, line 503:
```suggestion
def test_build_graph_to_dask_combine() -> None:
```

src/sciline/pipeline.py
SimonHeybrock: Do you mind just getting `RecursionError`?

tests/pipeline_test.py
```diff
@@ -483,3 +481,42 @@ def func(x: int, y: float) -> str:
...
+        nonlocal ncall
+        ncall += 1
+        return 3
+
+    pipeline = sl.Pipeline([int_to_float, provide_int, int_float_to_str])
+    graph1 = pipeline.build(float)
+    graph2 = pipeline.build(str)
+    graph = {**graph1, **graph2}
+
+    from sciline.pipeline import as_dask_graph
```
nvaytet, line 516: This is already imported at the top of the file I think?

nvaytet (commented):
Some small comments. Looks good.

src/sciline/graph.py
SimonHeybrock: Sure, the main point of this PR is the other refactor.

src/sciline/pipeline.py
jl-wynen: Yes. It would be nicer to report cycles in the graph more clearly. Maybe even during construction of the pipeline.

SimonHeybrock:
> I'd even have a separate graph class that only has the encoding and graph primitives in addition to a task graph class that has the higher level ops like compute

This is essentially what a dask collection is (so original PR message). I am not sure this is the right interface for us, since we may frequently want to compute multiple intermediate results. Will need more thought before moving on.

jl-wynen:
> > I'd even have a separate graph class that only has the encoding and graph primitives in addition to a task graph class that has the higher level ops like compute
> 
> This is essentially what a dask collection is (so original PR message). I am not sure this is the right interface for us, since we may frequently want to compute multiple intermediate results. Will need more thought before moving on.

Not really. A dask collection is more like a node of the graph (with links to parents). So it defines a graph as a linked structure. I was talking about a graph class that has a useful encoding for the graph plus methods like `dfs`, `add_node`, etc. And a task graph class that knows how to call these methods to string together computations to get a certain result or make a dask graph.
The reason is that this lets us abstract away the graph fro m the rest. If this becomes complicated enough, we may want to switch to a library like networkx instead of implementing graph algorithms ourselves.

SimonHeybrock:
> > > I'd even have a separate graph class that only has the encoding and graph primitives in addition to a task graph class that has the higher level ops like compute
> > 
> > 
> > This is essentially what a dask collection is (so original PR message). I am not sure this is the right interface for us, since we may frequently want to compute multiple intermediate results. Will need more thought before moving on.
> 
> Not really. A dask collection is more like a node of the graph (with links to parents). So it defines a graph as a linked structure.

Not really. A dask collection is a graph (given by a dict), and a set of keys.

> I was talking about a graph class that has a useful encoding for the graph plus methods like `dfs`, `add_node`, etc. And a task graph class that knows how to call these methods to string together computations to get a certain result or make a dask graph.

My initial comment was on the task graph class, not the graph class

SimonHeybrock:
Due you guys prefer merging this, or do you want to see two more days of follow-up changes in the same PR?

jl-wynen (approved):
> Due you guys prefer merging this, or do you want to see two more days of follow-up changes in the same PR?

I'm pretty sure we'll go with something like this. And since no one uses sciline, you can merge it. Let's keep the diffs short.

PR #17, opened 2023-07-18
Remove dask dependency and add new test run with dask

SimonHeybrock:
- Make dask optional.
- Add new test env with dask.


src/sciline/pipeline.py
```diff
@@ -239,6 +239,11 @@ def __init__(
     ) -> None:
         self._graph = graph
         self._keys = keys
+        if scheduler is None:
+            try:
+                scheduler = DaskScheduler()
+            except ImportError:
+                scheduler = NaiveScheduler()
         self._scheduler = scheduler or DaskScheduler()
```
YooSunYoung, line 247: Is this line needed? Shouldn't it be just 
`self._scheduler = scheduler`
??

YooSunYoung: I just found that it is updated in the next PR.

PR #18, opened 2023-07-18
Update and add tests

SimonHeybrock:
Also fixes the lack of support for `sc.Scope` subclasses when using instances as "providers".

### DIFF
--- a/tests/complex_workflow_test.py
+++ b/tests/complex_workflow_test.py
@@ -11,16 +11,16 @@ import sciline as sl
 
 @dataclass
 class RawData:
-    data: npt.NDArray[np.float64]
+    data: np.ndarray
     monitor1: float
     monitor2: float
 
 
 SampleRun = NewType('SampleRun', int)
 BackgroundRun = NewType('BackgroundRun', int)
-DetectorMask = NewType('DetectorMask', npt.NDArray[np.float64])
-DirectBeam = NewType('DirectBeam', npt.NDArray[np.float64])
-SolidAngle = NewType('SolidAngle', npt.NDArray[np.float64])
+DetectorMask = NewType('DetectorMask', np.ndarray)
+DirectBeam = NewType('DirectBeam', np.ndarray)
+SolidAngle = NewType('SolidAngle', np.ndarray)
 
 Run = TypeVar('Run')
 
@@ -29,7 +29,7 @@ class Raw(sl.Scope[Run], RawData):
     ...
 
 
-class Masked(sl.Scope[Run], npt.NDArray[np.float64]):
+class Masked(sl.Scope[Run], np.ndarray):
     ...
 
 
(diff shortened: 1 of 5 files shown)


tests/complex_workflow_test.py
```diff
@@ -80,26 +80,13 @@ def iofq(
...
-
-def direct_beam() -> DirectBeam:
-    return DirectBeam(np.array(1 / 1.5))
+params = {
+    Raw[SampleRun]: Raw(RawData(data=np.ones(4), monitor1=1.0, monitor2=2.0)),
+    Raw[BackgroundRun]: Raw(RawData(data=np.ones(4) * 1.5, monitor1=1.0, monitor2=4.0)),
+    DetectorMask: DetectorMask(np.array([1, 1, 0, 1])),
+    SolidAngle: SolidAngle(np.array([1.0, 0.5, 0.25, 0.125])),
+    DirectBeam: DirectBeam(np.array(1 / 1.5)),
+}
```
YooSunYoung, line 89: I like this way much better...!

tests/pipeline_test.py
```diff
@@ -58,7 +60,47 @@ def provide_int() -> int:
     assert pipeline.compute((float, str)) == (1.5, "3;1.5")
```
YooSunYoung, line 60: Shouldn't it also have
``assert ncall == 1``
??

SimonHeybrock, line 60:
```suggestion
    assert pipeline.compute((float, str)) == (1.5, "3;1.5")
    assert ncall == 1
```

tests/pipeline_test.py
```diff
@@ -21,19 +21,21 @@ def int_float_to_str(x: int, y: float) -> str:
...
     pipeline = sl.Pipeline([int_to_float])
     with pytest.raises(sl.UnsatisfiedRequirement):
         pipeline.compute(float)
 
 
-def test_intermediate_computed_once() -> None:
+def test_intermediate_used_multiple_times_is_computed_only_once() -> None:
     ncall = 0
 
     def provide_int() -> int:
```
YooSunYoung, line 41: Do you maybe want to use `unittest.mock` for counting calls?

Something like this...? Then we can use the call counting method of `Mock`.
```
from unittest.mock import Mock
MockCounter = NewType('MockCounter', Mock)
def provide_int(counter: MockCounter) -> int:
    counter()
    return 3

def test_whatever() -> None:
    mock_counter = Mock(lambda: None)
    pipeline = sl.Pipeline([int_to_float, provide_int, int_float_to_str], params={MockCounter: mock_counter})
    mock_counter.assert_called_once()
```

SimonHeybrock: We can consider this if we begin to add more complex tests. For now, the manual way seems sufficient?

YooSunYoung (approved):

PR #19, opened 2023-07-19
Avoid recursion and allow building cyclic graphs

SimonHeybrock:
- Avoid recursion. Not sure we ever get to graphs of a size where this would matter, but still.
- Allow building graphs with cycles. Main motivation was to be able to visualize them, which could be valuable for debugging.


src/sciline/pipeline.py
```diff
@@ -191,20 +190,22 @@ def build(self, tp: Type[T], /) -> Graph:
...
-            if name != 'return'
-        }
-        graph: Graph = {tp: (provider, args)}
-        try:
+        graph: Graph = {}
+        stack: List[Tuple[Type[T], Dict[TypeVar, Key]]] = [(tp, {})]
+        while stack:
+            tp, bound = stack.pop()
+            provider: Callable[..., T]
+            provider, bound = self._get_provider(tp)
```
YooSunYoung, line 198: If `bound` is going to be retrieved by `self._get_provider`
`stack` doesn't need to remember `bound`, right...?

tests/pipeline_test.py
```diff
@@ -492,16 +491,27 @@ class A(sl.Scope[T], int):
...
+def test_computing_graph_with_cycle_raises_CycleError() -> None:
+    def f(x: int) -> float:
+        return float(x)
+
+    def g(x: float) -> int:
+        return int(x)
+
+    pipeline = sl.Pipeline([f, g])
+    with pytest.raises(sl.scheduler.CycleError):
+        pipeline.compute(int)
```
YooSunYoung, line 514: Just a question,
Will it use `DaskScheduler` in `tox -e dask` and `NaiveScheduler` in `tox -e py38`...?

SimonHeybrock: Yes.

src/sciline/pipeline.py
SimonHeybrock: Good point, fixed!

YooSunYoung (approved):

PR #21, opened 2023-07-19
Getting started docs

SimonHeybrock:


docs/user-guide/getting-started.ipynb
```diff
@@ -0,0 +1,237 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Getting Started"
+   ]
+  },
```
YooSunYoung, line 10: How about adding a brief explanation about how `sciline.Pipeline` works...?
Mentioning it is mainly using type hints for connecting each steps into one big pipeline(workflow).

For example,

--------------------
## What is ``sciline``?
``sciline`` is a static-type based workflow development framework.
It collects dependencies based on the annotated types of arguments and return type of callable objects and compute the desired result.
Therefore each step of workflow are independent from other steps, which allows much more flexible development of workflow as well as testing of each step.

SimonHeybrock: I have something similar in https://github.com/scipp/sciline/pull/24 (on the front page), but will look into extending this here as well.

SimonHeybrock: Which parts of what you are suggesting are missing from the second subsection?

YooSunYoung: I thought it might be better to have `## What is sciline` section before `## Motivation`.
So that a reader can easily know why this `motivation` drove us to that solution mentioned earlier.

SimonHeybrock: Hmm, but usually the motivation comes first doesn't it?

SimonHeybrock: On second though: Thinking of a scientific paper as an analogy, I was thinking of the motivation as the first section (such as introduction). What you suggest might be seen as an abstract, which just summarizes briefly?

YooSunYoung: Yes that sounds more correct, that we need an abstract.
I just saw the #24 `## Why Sciline`, but it's more of a summary of this motivation section I thought.
So what I suggest is to add a brief description of the methodology, which is type annotations.

YooSunYoung (approved):

PR #23, opened 2023-07-25
Add first logo version

SimonHeybrock:


resources/sciline.svg
YooSunYoung, line 1: Why does this one not have an `e` at the end...?

SimonHeybrock: It does, but the page size is not correct, so the GitHub preview does not show it correctly.

YooSunYoung (approved):

PR #24, opened 2023-07-25
Add API Reference docs page and improve docstrings

SimonHeybrock:


src/sciline/domain.py
```diff
@@ -7,6 +7,10 @@
 
 
 class Scope(Generic[PARAM, SUPER]):
+    """
+    Helper for defining a generic type alias.
+    """
```
YooSunYoung, line 12: Should we maybe add examples here...? Including what is `PARAM` and `SUPER`.
It seems like type-argument parameters aren't really shown in the api reference.

SimonHeybrock: I am just trying to figure out where to document this. I will most likely add a notebook with examples.

SimonHeybrock: ... but will probably also wait until we have tried this more in practice, since I suspect the implementation may change a couple more times.

YooSunYoung (approved):

PR #25, opened 2023-07-26
Add mypy CI job

SimonHeybrock:


.github/workflows/ci.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-20.04']
-        tox-env: [py38, dask]
+        tox-env: [py38, dask, mypy]
```
YooSunYoung, line 33: Will it be enough to run mypy on 3.8...?

YooSunYoung (approved):

.github/workflows/ci.yml
SimonHeybrock: Indeed, we will probably want to add more in the future?

PR #28, opened 2023-07-26
Fix duplicate display of generic params in visualize()

SimonHeybrock:
This was probably broken by unrelated changes to how the graph was stored.


YooSunYoung:
Will it be confusing If there are different domain types with same names...?
If they are defined in different modules, the `Pipeline` will not be confused since they are different objects,
but if we only use the `__name__` then it doesn't show which module they are from.

PR #33, opened 2023-08-02
Add support for parameter tables

### LINKED ISSUE #32: Variadic workflows
There can be cases where a workflow has to deal with an unknown number of inputs, such as loading an unknown number of files, which are then combined.

One way of handling this would be to define domain types such as `FilenameList` and `ImageList`, as well as associated providers. However, this can quickly become tedious. Can we handle this better, something like https://injector.readthedocs.io/en/latest/api.html#injector.Binder.multibind, but also on the provider level?

SimonHeybrock:
Fixes #32.

Support parameter tables, enabling "map"-, "reduce"-, and "groupby"-like graph building. Please see the added documentation page in the user guide for details.

Internal change: The graph now does not use keyword arguments any more. This did not work with Dask anyway, and just got in the way when adding the new functionality.


src/sciline/param_table.py
```diff
@@ -0,0 +1,86 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from __future__ import annotations
+
+from typing import Any, Collection, Dict, Mapping, Optional
```
jl-wynen, line 5: `typing.Collection` is deprecated in favour of `collections.abc.Collection` in python 3.9

src/sciline/pipeline.py
```diff
@@ -39,8 +55,12 @@ class AmbiguousProvider(Exception):
     """Raised when multiple providers are found for a type."""
 
 
-Provider = Callable[..., Any]
-Key = type
+def _indexed_key(index_name: Any, i: int, value_name: Type[T] | Item[T]) -> Item[T]:
```
jl-wynen, line 58: I thought you want to support python 3.8? This use of the pipe was added in 3.10

src/sciline/pipeline.py
```diff
@@ -74,14 +94,139 @@ def _bind_free_typevars(tp: TypeVar | Key, bound: Dict[TypeVar, Key]) -> Key:
...
+        return []
+    paths = []
+    for node in graph[start]:
+        for path in _find_all_paths(graph, node, end):
+            paths.append([start] + path)
+    return paths
+
+
+def _find_nodes_in_paths(
+    graph: Mapping[T, Tuple[Callable[..., Any], Collection[T]]], start: T, end: T
```
jl-wynen, line 113: This function and `_find_all_paths` use different graph types. Can you give the arguments a clearer name to distinguish? The type hint alone doesn't really say what kind of graph we are dealing with.

Also, why are there different graph types in the first place?

src/sciline/pipeline.py
```diff
@@ -74,14 +94,139 @@ def _bind_free_typevars(tp: TypeVar | Key, bound: Dict[TypeVar, Key]) -> Key:
...
+) -> List[T]:
+    """
+    Find all nodes that need to be duplicated since they depend on a value from a
+    param table.
```
jl-wynen, line 117: The docstring refers to a specific use case, not to what the function does.

src/sciline/pipeline.py
```diff
@@ -153,10 +300,48 @@ def __setitem__(self, key: Type[T], param: T) -> None:
...
+        A pipeline can have multiple parameter tables, but only one per row dimension.
+        Column names must be unique across all parameter tables.
+
+        Parameters
+        ----------
+        params:
+            Parameter table to set.
+        """
+        if params.row_dim in self._param_tables:
+            raise ValueError(f'Parameter table for {params.row_dim} already set')
```
jl-wynen, line 319: What if a user wants to replace a table? Is this a reasonable thing to want to do?

src/sciline/pipeline.py
```diff
@@ -96,6 +241,8 @@ def __init__(
         """
         self._providers: Dict[Key, Provider] = {}
         self._subproviders: Dict[type, Dict[Tuple[Key | TypeVar, ...], Provider]] = {}
+        self._param_tables: Dict[Key, ParamTable] = {}
+        self._param_series: Dict[Key, Key] = {}
```
jl-wynen, line 245: Confusing name. This maps from parameter names to dimensions., not so series objects.

src/sciline/pipeline.py
```diff
@@ -153,10 +300,48 @@ def __setitem__(self, key: Type[T], param: T) -> None:
...
+        self, key: Union[Type[T], Item[T]], provider: Callable[..., T]
+    ) -> None:
         # isinstance does not work here and types.NoneType available only in 3.10+
         if key == type(None):  # noqa: E721
             raise ValueError(f'Provider {provider} returning `None` is not allowed')
+        if get_origin(key) == Series:
+            raise ValueError(
+                f'Provider {provider} returning a sciline.Series is not allowed. '
+                'Series is a special container reserved for use in conjunction with '
+                'sciline.ParamTable and may not be provided directly.'
```
jl-wynen, line 343:
```suggestion
                'sciline.ParamTable and must not be provided directly.'
```

src/sciline/pipeline.py
```diff
@@ -206,25 +395,111 @@ def build(self, tp: Type[T], /) -> Graph:
...
-            }
+            )
             graph[tp] = (provider, args)
-            for arg in args.values():
+            for arg in args:
                 if arg not in graph:
                     stack.append(arg)
         return graph
 
+    def _build_series(self, tp: Type[Series[KeyType, ValueType]]) -> Graph:
```
jl-wynen, line 425: Quite a long and deeply nested function. Can you split it up, e.g., along the steps?

Also, I don't understand it. We may need to go through it together. But refactoring may already help.

src/sciline/pipeline.py
```diff
@@ -206,25 +395,111 @@ def build(self, tp: Type[T], /) -> Graph:
...
+        # a Series, this will terminate when it reaches a parameter that is not a
+        # single provided value but a collection of values from a parameter table
+        # column. Instead of single value (which does not exist), a sentinel is
+        # used to mark this, for processing below.
+        subgraph = self.build(value_type, search_param_tables=True)
+        # Step 2:
+        # Identify nodes in the graph that need to be duplicated as they lie in the
+        # path to a parameter from a table. In the case of grouping, note that the
+        # ungrouped graph (including duplication of nodes) will have been built by a
+        # prior call to _build_series, so instead of duplicated everything until the
```
jl-wynen, line 440:
```suggestion
        # prior call to _build_series, so instead of duplicating everything until the
```

tests/graph_test.py
```diff
@@ -0,0 +1,10 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from sciline.pipeline import _find_all_paths
+
+
+def test_find_all_paths() -> None:
+    graph = {"D": ["B", "C"], "C": ["A"], "B": ["A"]}
+    assert _find_all_paths(graph, "D", "A") == [["D", "B", "A"], ["D", "C", "A"]]
+    assert _find_all_paths(graph, "B", "C") == []
+    assert _find_all_paths(graph, "B", "A") == [["B", "A"]]
```
jl-wynen, line 10: Missing cases for an empty graph.
But this tests an implementation detail. Can we do without it?

jl-wynen (commented):
Incomplete review

src/sciline/param_table.py
SimonHeybrock: My problem is that `typing.abc.* is not generic in 3.8.

src/sciline/pipeline.py
SimonHeybrock: Hmm, odd, why is it passing builds then?

src/sciline/pipeline.py
SimonHeybrock: We are currently not allowing this for providers and params either. Could be added in the future, but maybe not worth the complexity.

src/sciline/pipeline.py
SimonHeybrock: Yes, this is the part that I wanted to do the walk-through for.

src/sciline/pipeline.py
SimonHeybrock: Yes, is that bad? It is an implementation detail.

tests/graph_test.py
SimonHeybrock: Probably, I think I just added it during dev.

src/sciline/pipeline.py
jl-wynen: It confused my at first. Maybe just mention that this is the particular use case.

src/sciline/param_table.py
```diff
@@ -0,0 +1,86 @@
...
+        """The row dimension of the table."""
+        return self._row_dim
+
+    @property
+    def index(self) -> Collection[Any]:
+        """The row index of the table."""
+        return self._index
+
+    def __contains__(self, key: Any) -> bool:
+        return self._columns.__contains__(key)
```
jl-wynen, line 62: Why did you use the dunder methods of `self._columns` instead of the builtin functions which call them?

src/sciline/visualize.py
```diff
@@ -1,53 +1,97 @@
...
+        If True, parameter-table-dependent branches are collapsed into a single copy
+        of the branch. Recommendend for large graphs with long parameter tables.
     kwargs:
         Keyword arguments passed to :py:class:`graphviz.Digraph`.
     """
     dot = Digraph(strict=True, **kwargs)
-    for p, (p_name, args, ret) in _format_graph(graph).items():
-        dot.node(ret, ret, shape='rectangle')
+    for p, (p_name, args, ret) in _format_graph(graph, compact=compact).items():
+        if '(' in ret and '=' not in ret:
```
jl-wynen, line 38: This seems brittle. Can you return an object with metadata from `_format_graph` instead of plain strings?

src/sciline/visualize.py
```diff
@@ -1,53 +1,97 @@
...
+                dot.edge(arg, p)
+            dot.edge(p, ret)
     return dot
 
 
-def _format_graph(graph: Graph) -> Dict[str, Tuple[str, List[str], str]]:
+def _qualname(obj: Any) -> Any:
+    return (
+        obj.__qualname__ if hasattr(obj, '__qualname__') else obj.__class__.__qualname__
+    )
```
jl-wynen, line 64:
```suggestion
def _qualname(obj: Any) -> Any:
    return getattr(obj, '__qualname__', obj.__class__.__qualname__)
```

tests/pipeline_with_index_test.py
jl-wynen, line 1: Rename the file? I don't know what 'index' refers to here.

tests/pipeline_with_index_test.py
```diff
@@ -0,0 +1,570 @@
...
+    def use_str(x: str) -> Str:
+        return Str(x)
+
+    pl = sl.Pipeline([gather_zip, use_str])
+    pl.set_param_table(sl.ParamTable(Run, {str: ['a', 'a', 'ccc'], int: [1, 2, 3]}))
+
+    assert pl.compute(Sum) == "['a1', 'a2', 'ccc3']"
+
+
+def test_diamond_dependency_pulls_values_from_columns_in_same_param_table() -> None:
```
jl-wynen, line 131: This does not look like a diamond dependency to me. It looks like a Y: Sum depends on float which depends on Param1 and Param2. But those depend on nothing.

src/sciline/param_table.py
SimonHeybrock: No reason.

src/sciline/visualize.py
SimonHeybrock: Won't that raise if `obj` has no `__class__`?

tests/pipeline_with_index_test.py
SimonHeybrock: I meant: they both depend on the same param table and thus index.

src/sciline/visualize.py
jl-wynen: True. Leave it as is

src/sciline/pipeline.py
SimonHeybrock: I completely rewrote this, moving the functionality mostly into the "grouper" classes. I think it is more readable now, please have a look @jl-wynen .

src/sciline/pipeline.py
```diff
@@ -206,25 +448,123 @@ def build(self, tp: Type[T], /) -> Graph:
...
+        build(Series[Idx,B]) will call _build_series(Series[Idx,B]). This results in
+        the following procedure here:
+
+        1. Call build(P1), resulting, e.g., in a graph S->A->B, where S is a sentinel.
+           The sentinel is used because build() cannot find a unique P1, since it is
+           not a single value but a column in a table.
+        2. Instantiation of `Replicator`, which will be used to replicate the
+           relevant parts of the graph (see illustration there).
+        3. Insert a special `SeriesProvider` node, which will gather the duplicates of
+           the 'B' node and providers the requested Series[Idx,B].
```
jl-wynen, line 499:
```suggestion
           the 'B' node and provides the requested Series[Idx,B].
```

jl-wynen (approved):

PR #35, opened 2023-08-10
Clarify scope / discuss Snakemake

SimonHeybrock:


docs/index.md
```diff
@@ -11,10 +11,33 @@ Writing, testing, and maintaining complex data analysis workflows is hard.
...
+
+### Why not Snakemake?
+
+Sciline is not intended to replace Snakemake or similar systems.
+We see Snakemake as a complementary tool that operates on a higher level.
+
+### But I do not want to change all my code!
+
+Sciline is very non-invasive.
+If you are willing to define a few types (or type aliases) and add [type](https://docs.python.org/3/library/typing.html) [annotations](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html) to your functions &mdash; which is considered good practice anyway &mdash; you can use it with Sciline with just a few extra lines of code.
```
jl-wynen, line 38: I wouldn't say that adding lots of custom types is good practice in general. But adding type hints is.

jl-wynen, line 38: Nice idea with the two links but this will render in a non-obvious way because they are right next to each other. I'd add an explicit link to the cheat sheet.

jl-wynen (approved):
Looks good. But I'm biased because I know Sciline from the inside.

docs/index.md
SimonHeybrock:
```suggestion
If you are willing to define a few types (or type aliases) and add [type annotations](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html) to your functions &mdash; which is considered good practice anyway &mdash; you can use it with Sciline with just a few extra lines of code.
```

I'll keep only the cheat-sheet link.

SimonHeybrock: I agree with you, but it feels like nitpicking in this context, I think it is enough for the purpose of a brief FAQ entry.

PR #44, opened 2023-08-31
``compute`` returns ``dict`` instead of ``tuple``.

### LINKED ISSUE #27: Return `dict` when computing multiple results
> I have to correct myself here: If we use an alias via `NewType` or `Scope` then the instance we get as a result does not have that info. Therefore, I think you are right and we should return a `dict`.

_Originally posted by @SimonHeybrock in https://github.com/scipp/sciline/pull/16#discussion_r1275024564_

```python
# Now
result = pipeline.compute((IofQ, RawData))
result[1]
# Then?
result = pipeline.compute_dict((IofQ, RawData))
result[RawData]
```

YooSunYoung:
Fixes #27

### DIFF
--- a/src/sciline/task_graph.py
+++ b/src/sciline/task_graph.py
@@ -37,7 +37,7 @@ class TaskGraph:
     def compute(
         self,
         keys: Optional[
-            Union[type, Tuple[type, ...], Item[T], Tuple[Item[T], ...], None]
+            Union[type, Tuple[type, ...], Item[T], Tuple[Item[T], ...]]
         ] = None,
     ) -> Any:
         """
--- a/src/sciline/task_graph.py
+++ b/src/sciline/task_graph.py
@@ -49,23 +49,19 @@ class TaskGraph:
             Optional list of keys to compute. This can be used to override the keys
             stored in the graph instance. Note that the keys must be present in the
             graph as intermediate results, otherwise KeyError is raised.
-        
+
         Returns
         -------
         If ``keys`` is a single type, returns the single result that was computed.
         If ``keys`` is a tuple of types, returns a dictionary with type as keys
-        and the coresponding results as values.
+        and the corresponding results as values.
 
         """
         if keys is None:
             keys = self._keys
         if isinstance(keys, tuple):
             results = self._scheduler.get(self._graph, list(keys))
-            return {
-                tp: result
-                for tp, result
-                in zip(keys, results)
-            }
+            return {tp: result for tp, result in zip(keys, results)}
         else:
             return self._scheduler.get(self._graph, [keys])[0]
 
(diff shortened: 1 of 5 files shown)


src/sciline/task_graph.py
```diff
@@ -37,7 +37,7 @@ def __init__(
     def compute(
         self,
         keys: Optional[
-            Union[type, Tuple[type, ...], Item[T], Tuple[Item[T], ...]]
+            Union[type, Tuple[type, ...], Item[T], Tuple[Item[T], ...], None]
```
SimonHeybrock, line 40: Why the `None` here?

src/sciline/task_graph.py
```diff
@@ -49,11 +49,19 @@ def compute(
             Optional list of keys to compute. This can be used to override the keys
             stored in the graph instance. Note that the keys must be present in the
             graph as intermediate results, otherwise KeyError is raised.
+
+        Returns
+        -------
+        If ``keys`` is a single type, returns the single result that was computed.
+        If ``keys`` is a tuple of types, returns a dictionary with type as keys
+        and the corresponding results as values.
```
SimonHeybrock, line 57: As @jl-wynen pointed out: mypy will not be happy? Does adding an `@overload` help?

src/sciline/task_graph.py
```diff
@@ -49,11 +49,19 @@ def compute(
...
+
         """
         if keys is None:
             keys = self._keys
         if isinstance(keys, tuple):
-            return self._scheduler.get(self._graph, list(keys))
+            results = self._scheduler.get(self._graph, list(keys))
+            return {tp: result for tp, result in zip(keys, results)}
```
SimonHeybrock, line 64:
```suggestion
            return dict(zip(keys, results))
```

src/sciline/task_graph.py
YooSunYoung: It is `Any` so it shouldn't matter, no...?

src/sciline/task_graph.py
SimonHeybrock: Why do you prefer that?

src/sciline/task_graph.py
SimonHeybrock: True. Can we split it into `Any` and `Dict[type, Any]` now?

src/sciline/task_graph.py
YooSunYoung: It doesn't really matter since they are basically the same.
Just less lines... I guess...? I'll revert it anyways : D

src/sciline/task_graph.py
YooSunYoung: You mean like this...?
```python
def compute(self, tp: type | Tuple[type, ...] | Item[T]) -> Any | Dict[Type[T], T]: ...
```

YooSunYoung: Isn't it already like this in `pipeline.Pipeline`? and `task_graph.TaskGraph` can also have them but it doesn't really work with `None` since the returned type depends on the `self._keys` attribute type, not the argument type.

SimonHeybrock: Right.

PR #50, opened 2023-09-01
Better handling of `typing.Optional`

### LINKED ISSUE #40: Handling of `typing.Optional`
Consider a function

```python
def func(x: List[int], param: Optional[str] = None): ...
```
How should Sciline handle the `Optional[str]` parameter,

1. look for `str` and inject it if found, or
2. look for `Optional[str]` and inject it, or
3. both?

- Option 1.) would be useful when `str` is provided as a parameter or from a parameter table.
- Option 2.) would be useful if there is a provider
  ```python
  def maybe_str() -> Optional[str]: ...
  ```

SimonHeybrock:
Fixes #40.

Should we add top-level docs for this?

### DIFF
--- a/tests/pipeline_with_optional_test.py
+++ b/tests/pipeline_with_optional_test.py
@@ -83,6 +83,21 @@ def test_optional_dependency_is_set_to_none_if_no_provider_found_transitively():
     assert pipeline.compute(str) == '123'
 
 
+def test_optional_dependency_in_node_depending_on_param_table():
+    def use_optional(x: float, y: Optional[int]) -> str:
+        return f'{x} {y or 123}'
+
+    pl = sl.Pipeline([use_optional])
+    pl.set_param_table(sl.ParamTable(int, {float: [1.0, 2.0, 3.0]}))
+    assert pl.compute(sl.Series[int, str]) == sl.Series(
+        int, {0: '1.0 123', 1: '2.0 123', 2: '3.0 123'}
+    )
+    pl[int] = 11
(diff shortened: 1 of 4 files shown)


tests/pipeline_with_optional_test.py
```diff
@@ -0,0 +1,140 @@
...
+        pipeline[Union[int, float]] = 3  # type: ignore[index]
+    with pytest.raises(ValueError):
+        pipeline[Optional[int]] = 3  # type: ignore[index]
+
+
+def test_union_requirement_leads_to_UnsatisfiedRequirement() -> None:
+    def require_union(x: Union[int, float]) -> str:
+        return f'{x}'
+
+    pipeline = sl.Pipeline([require_union])
```
jl-wynen, line 38: Can you add a provider for `int` and/or `float` so this actually tests something?

tests/pipeline_with_optional_test.py
jl-wynen, line 1: Please add tests where there is an optional and a required dependency on some value. And check that it works for any order or providers and whether the value can be provided or not.

SimonHeybrock: I added one test, not entirely sure what the other cases you are asking for are.

tests/pipeline_with_optional_test.py
SimonHeybrock: Good point, see update!

jl-wynen (approved):

PR #51, opened 2023-09-05
Cluster generic products when visualizing graph

SimonHeybrock:
Without this, graphs that make heavy use of generics are pretty unreadable. It is enabled by default.


jl-wynen:
Can you post images with and without this option?

SimonHeybrock:
@jl-wynen 
image

SimonHeybrock:
One more, in top-down layout:

image
image

jl-wynen (approved):

PR #53, opened 2023-09-06
Add ScopeTwoParams

### LINKED ISSUE #42: `sciline.Scope` with multiple type params
To avoid code duplication, we would like to be able to do:

```python
class RawMonitor(sciline.Scope[RunType, MonitorType, sc.DataArray], sc.DataArray):
    ...
```

SimonHeybrock:
Fixes #42.

Hope to be able to find a nicer solution once we can use TypeVarTuple.

### DIFF
--- a/src/sciline/domain.py
+++ b/src/sciline/domain.py
@@ -3,7 +3,32 @@
 from typing import Any, Generic, TypeVar, get_args, get_origin
 
 PARAM = TypeVar("PARAM")
+PARAM2 = TypeVar("PARAM2")
 SUPER = TypeVar("SUPER")
+
+
+def _check_supertype(cls: type) -> None:
+    # Mypy does not support __orig_bases__ yet(?)
+    # See also https://stackoverflow.com/a/73746554 for useful info
+    scope = cls.__orig_bases__[0]  # type: ignore[attr-defined]
+    # Only check direct subclasses
+    if get_origin(scope) is Scope:
+        supertype = get_args(scope)[1]
+        # Remove potential generic params
+        # In Python 3.8, get_origin does not work with numpy.typing.NDArray,
+        # but it defines __origin__
+        supertype = getattr(supertype, '__origin__', None) or supertype
+        if supertype not in cls.__bases__:
+            raise TypeError(
+                f"Missing or wrong interface for {cls}, "
+                f"should inherit {supertype}.\n"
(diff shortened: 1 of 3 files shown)


src/sciline/domain.py
```diff
@@ -3,36 +3,55 @@
...
 SUPER = TypeVar("SUPER")
 
 
+def _check_supertype(cls: type) -> None:
+    # Mypy does not support __orig_bases__ yet(?)
+    # See also https://stackoverflow.com/a/73746554 for useful info
+    scope = cls.__orig_bases__[0]  # type: ignore[attr-defined]
+    # Only check direct subclasses
+    if get_origin(scope) is Scope:
+        supertype = get_args(scope)[1]
```
YooSunYoung, line 16: Shouldn't it be the last one of the type arguments?

```python
supertype = get_args(scope)[-1]
```

SimonHeybrock: Good catch! Also the line above was wrong. Added a test.

src/sciline/domain.py
```diff
@@ -3,36 +3,55 @@
...
+    # Mypy does not support __orig_bases__ yet(?)
+    # See also https://stackoverflow.com/a/73746554 for useful info
+    scope = cls.__orig_bases__[0]  # type: ignore[attr-defined]
+    # Only check direct subclasses
+    if get_origin(scope) is scope_cls:
+        supertype = get_args(scope)[-1]
+        # Remove potential generic params
+        # In Python 3.8, get_origin does not work with numpy.typing.NDArray,
+        # but it defines __origin__
+        supertype = getattr(supertype, '__origin__', None) or supertype
```
YooSunYoung, line 20: Is it assuming any of `Scope` classes does not have `__origin__` defined...?

YooSunYoung: No sorry I was a bit confused... why is it getting `__origin__` of the `supertype`...?
Can you tell me a use-case...? 🥲

SimonHeybrock: If you make an alias of a generic type, this is needed for the `__bases__` check below.

Note that this is just moved, i.e., not new in this PR.

YooSunYoung: Ah right now I understood. Thank you!

YooSunYoung (approved):

PR #55, opened 2023-09-11
Add recipe for file output and side effects

### LINKED ISSUE #29: Handling file save or upload operations
Sciline is based on requesting a specific output (type). When we save a file or upload a file, then this would naively be `None`, which would not work. We can work around this, e.g., by returning a status or success object from the provider, which the user can "request".

There may be nothing to do here (in code), but we should probably write a documentation page with a useful suggestion, since I imagine this may not be obvious to users.

jl-wynen:
Fixes #29

### DIFF
--- a/docs/user-guide/recipes.ipynb
+++ b/docs/user-guide/recipes.ipynb
@@ -9,18 +9,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "fce21908-20af-4580-afc7-08aaa152e828",
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "from typing import NewType\n",
-    "\n",
-    "import sciline"
-   ]
-  },
-  {
    "cell_type": "markdown",
    "id": "0835a63d-f22a-477e-b887-9cdba171d8f0",
    "metadata": {},
@@ -51,6 +39,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from typing import NewType\n",
+    "\n",
+    "import sciline\n",
+    "\n",
     "_fake_filesystem = {}\n",
     "\n",
     "Param = NewType('Param', float)\n",
--- a/docs/user-guide/recipes.ipynb
+++ b/docs/user-guide/recipes.ipynb
@@ -25,7 +25,7 @@
    "id": "0835a63d-f22a-477e-b887-9cdba171d8f0",
    "metadata": {},
    "source": [
-    "## Side effects\n",
+    "## Avoiding side effects\n",
     "\n",
     "It is strongly discouraged to use [side effects](https://en.wikipedia.org/wiki/Side_effect_%28computer_science%29) in code that runs as part of a pipeline.\n",
     "This applies to, among others, file output, setting global variables, or communicating over a network.\n",
@@ -51,8 +51,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pathlib import Path\n",
-    "\n",
     "_fake_filesystem = {}\n",
     "\n",
     "Param = NewType('Param', float)\n",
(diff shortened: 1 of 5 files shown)


docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,139 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fce21908-20af-4580-afc7-08aaa152e828",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from typing import NewType\n",
+    "\n",
+    "import sciline"
```
SimonHeybrock, line 20: If you imagine adding more recipes, we should make each stand-alone, i.e., move imports into the concrete example.

docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,139 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0835a63d-f22a-477e-b887-9cdba171d8f0",
+   "metadata": {},
+   "source": [
+    "## Side effects\n",
```
SimonHeybrock, line 28:
```suggestion
    "## Avoiding side effects\n",
```

docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,139 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "d7608bb4-d7ce-4a36-8021-f1a52ad27ec5",
+   "metadata": {},
+   "source": [
+    "# Recipes"
```
SimonHeybrock, line 8: Should this be a sibling of `User Guide`, so it can be reached more easily in fewer clicks?

docs/user-guide/recipes.ipynb
```diff
@@ -0,0 +1,139 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8d39f143-c9da-4a7c-9ec8-ea7697b68b1b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from pathlib import Path\n",
```
SimonHeybrock, line 54: Unused? Did you plan to use it for the filename?

docs/user-guide/recipes.ipynb
jl-wynen: I figured that these two are needed everywhere. But I can change it.

docs/user-guide/recipes.ipynb
jl-wynen: Can be. We have limited real estate in the top bar. Do you think this is important / commonly used enough?

docs/user-guide/recipes.ipynb
jl-wynen: Right, I used real files initially, removing.

jl-wynen:
What do you think of the name 'Recipes'? What about 'Examples', 'How to', or something similar. This is the best time to change it...

docs/user-guide/recipes.ipynb
SimonHeybrock: Yes, but it makes it easier to copy-and-paste a working example if we repeat it. Or to split the page later.

docs/user-guide/recipes.ipynb
SimonHeybrock: I thought so, yes?

SimonHeybrock:
> What do you think of the name 'Recipes'? What about 'Examples', 'How to', or something similar. This is the best time to change it...

Recipes is good. Or "Recipes and Examples"?

docs/user-guide/recipes.ipynb
jl-wynen: Done

docs/user-guide/recipes.ipynb
jl-wynen: Done

jl-wynen:
I left it as recipes because that is short and the header is short, too.

SimonHeybrock (approved):

PR #65, opened 2023-10-16
Run copier update

SimonHeybrock:
Please review carefully, as copier overwrote some changes to the docs (which I caught when committing, but there may be more issues).


jl-wynen (approved):
I don't see any issues with the docs. The update seems quite straight forward. What problems did you have?

SimonHeybrock:
> I don't see any issues with the docs. The update seems quite straight forward. What problems did you have?

Copier overwrote changes to one of the `index.md` files.

PR #66, opened 2023-10-19
Run copier update

SimonHeybrock:


YooSunYoung:
Are we fixing all `txt` files in `requirements`?
I thought `.ini` files are enough to be in the template and each projects can fix their own `txt` files.

YooSunYoung (approved):

PR #67, opened 2023-10-23
fix: consider typevar constraints when finding providers

jokasimr:
Not sure if this is actually a bug or if the behavior is already as intended. In that case feel free to close.

**The problem**
This example
```python
import sciline
from typing import TypeVar, NewType

A = NewType('A', str)
B = NewType('B', str)
C = NewType('C', str)

V1 = TypeVar('V1', A, B)
V2 = TypeVar('V2', A, B, C)

class H(sciline.Scope[V2, str], str):
    pass

def p1(x: V1) -> H[V1]:
    return f"H[{x}]"

def p2(x: C) -> H[C]:
    return "H[C]"

sciline.Pipeline(
    [p1, p2], params={A: 'A', B: 'B', C: 'C'}
).visualize(H[C])
```
raises `AmbiguousProvider: ('Multiple providers found for type', __main__.H[__main__.C])`

I expected the provider would not be ambiguous because `C` is not in `V1`.
This PR makes the above case non-ambigous by taking into account the constraints on type vars when searching for compatible types.

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -53,21 +53,29 @@ class AmbiguousProvider(Exception):
     """Raised when multiple providers are found for a type."""
 
 
+def _is_compatible_type(
+    requested: Key,
+    provided: Key | TypeVar,
+) -> bool:
+    """
+    Check if a type is compatible to a provided type.
+    Types must either by equal, or the provided type must be a TypeVar
+    with constraints satisfied by the requested type."""
+    if isinstance(provided, TypeVar):
+        return len(provided.__constraints__) == 0 or any(
+            _is_compatible_type(requested, c) for c in provided.__constraints__
+        )
+    return requested == provided
+
+
 def _is_compatible_type_tuple(
     requested: tuple[Key, ...],
     provided: tuple[Key | TypeVar, ...],
 ) -> bool:
     """
     Check if a tuple of requested types is compatible with a tuple of provided types.
-
-    Types in the tuples must either by equal, or the provided type must be a TypeVar.
     """
-    for req, prov in zip(requested, provided):
-        if isinstance(prov, TypeVar):
-            continue
-        if req != prov:
-            return False
-    return True
+    return all(map(_is_compatible_type, requested, provided))
 
 
 def _bind_free_typevars(tp: TypeVar | Key, bound: Dict[TypeVar, Key]) -> Key:
--- a/tests/pipeline_test.py
+++ b/tests/pipeline_test.py
@@ -609,6 +609,30 @@ def test_init_with_sciline_Scope_subclass_param_works() -> None:
     assert pl.compute(A[str]) == A(2)
 
 
+def test_sciline_Scope_typevar_constraints_are_considered() -> None:
+    A = NewType('A', str)
+    B = NewType('B', str)
+    C = NewType('C', str)
+
+    V1 = TypeVar('V1', A, B)
+    V2 = TypeVar('V2', A, B, C)
+
+    class H(sl.Scope[V2, str], str):
+        pass
+
+    def p1(x: V1) -> H[V1]:
+        return H[V1](f"H[{x}]")
+
+    def p2(x: C) -> H[C]:
+        return H[C]("H[Clementine]")
+
+    pl = sl.Pipeline([p1, p2], params={A: 'A', B: 'B', C: 'C'})
+
+    assert str(pl.compute(H[A])) == "H[A]"
+    assert str(pl.compute(H[B])) == "H[B]"
+    assert str(pl.compute(H[C])) == "H[Clementine]"
+
+
 def test_building_graph_with_cycle_succeeds() -> None:
     def f(x: int) -> float:
         return float(x)


YooSunYoung:
I agree with the change!

Should we also have the failing test?? 
So that it elaborates that generic type providers with explicit type arguments can't overwrite generic type provider with ``TypeVar``.

Sth like:
```python
def test_sciline_overlapping_typevar_constaints_raises() -> None:
    A = NewType('A', str)
    B = NewType('B', str)

    V = TypeVar('V', A, B)

    class H(sl.Scope[V, str], str):
        pass

    def p1(x: V) -> H[V]:
        return H[V](f"H[{x}]")

    def p2(x: B) -> H[B]:
        return H[B]("H[Banana]")

    pl = sl.Pipeline([p1, p2], params={A: 'A', B: 'B'})

    assert str(pl.compute(H[A])) == "H[A]"
    with pytest.raises(sl.AmbiguousProvider):
        assert str(pl.compute(H[B]))

```

SimonHeybrock:
The behavior was "intended", in the sense that I deliberately did not support it initially, due to higher complexity. This does not mean we should not support it.

The use of constraints is nice, however I feel it might solve only a relatively low percentage of edge cases?

In practice, I think the most common use case would be a predefined workflow with generic providers. A specific `TypeVar` might be used in tens of places. If a user now wants to provide a specialization one of the providers, they cannot realistically go and edit most of the workflow (they might not even have access to it, if it is part of a package).

Therefore, given a provider `p`

```python
def p(x: V2) -> H[V2]:
    return f"H[{x}]"
```
 instead of being able to simply define `p2`, the user must also define `V1` and `p1` (forwarding to `p`). If `V2` in the package changes (e.g., adding another type), `V1` will be out of date and the pipeline is broken.

Is this too inconvenient?

The alternative would be to simply ignore constraints, favor specializations in the provider search, and fall back to generic providers if no specialization is found.

YooSunYoung:
> The alternative would be to simply ignore constraints, favor specializations in the provider search, and fall back to generic providers if no specialization is found.

I agree with prioritizing specific provider over generic providers,
but can we then not allow the types that are not in the constraints...?

e.g.
```python
S1 = NewType("S1", str)
S2 = NewType("S2", str)
S3 = NewType("S3", str)

S12 = TypeVar("T", S1, S2)

class H(sl.Scope[V, str], str):
    ...

def p12(S12) -> H[S12]
   ...

pl = sl.Pipeline(p12, params={S1: 'S1', S2:'S2', S3:'S3'})
pl.compute(H[S3])  # Should raise provider not found error

```

jokasimr:
My use case for this:
*  I have a provider (called `normalize`) that takes either a `Histogrammed[Sample]` or a `Histogrammed[CalibratedReference]` as input.
* The `Histogrammed[CalibratedReference]` cannot be provided by the same provider as `Histogrammed[Sample]`.
* The provider that produces `Histogrammed[Sample]` should also be able to produce a `Histogrammed[Reference]`.

This gives us the same structure as in the above example.

Alternative solution:
* Create a special type called  `HistogrammedCalibratedReference`.
* Drawback: Type nesting. If we create a special type then `normalize` would need to take a `TypeVar` being either a `Histogrammed[Sample]` or  a  `HistogrammedCalibratedReference` as input, and would produce the output types `Normalized[Histogrammed[Sample]]` and `Normalized[HistogrammedCalibratedReference]`.

SimonHeybrock:
> but can we then not allow the types that are not in the constraints...?

I am not sure this is beneficial? It would mean that providers defined in a package can never be used outside the strict scope defined by the package. Say my package defines a generic provider, but I need to add another case, or branch to a workflow. If we enforce constraints this will become impossible.

SimonHeybrock:
> My use case for this:
> 
>     * I have a provider (called `normalize`) that takes either a `Histogrammed[Sample]` or a `Histogrammed[CalibratedReference]` as input.
> 
>     * The `Histogrammed[CalibratedReference]` cannot be provided by the same provider as `Histogrammed[Sample]`.
> 
>     * The provider that produces `Histogrammed[Sample]` should also be able to produce a `Histogrammed[Reference]`.
> 
> 
> This gives us the same structure as in the above example.

Wouldn't it work by simply favoring specializations? Enforcing constraints does not appear essential for supporting this?

Enforcing constraints is something we should discussed, but it is at least partially independent of the "specialization" issue, since it would have much wider impact.

jokasimr:
Favoring specializations is a good idea.
How do we deal with "nested" `TypeVars`? Do we favor lower nesting over higher nesting?

SimonHeybrock:
> How do we deal with "nested" `TypeVars`? Do we favor lower nesting over higher nesting?

Can you give an example?

jokasimr:
Nesting like this:
```python
A = NewType('A', str)
TV1 = TypeVar('TV1', A, int)
TV2 = TypeVar('TV2', tv1, float)

def p1(a: TV1):
    return 'tv1[a]'

def p2(a: TV2):
    return 'tv2[a]'
```

The question was if we should favor `p1` over `p2`?

YooSunYoung:
> The question was if we should favor p1 over p2?

If we don't force any constraints, then I think they should be considered as ambiguous providers.

SimonHeybrock:
> Nesting like this:
> 
> ```python
> A = NewType('A', str)
> TV1 = TypeVar('TV1', A, int)
> TV2 = TypeVar('TV2', tv1, float)
> 
> def p1(a: TV1):
>     return 'tv1[a]'
> 
> def p2(a: TV2):
>     return 'tv2[a]'
> ```
> 
> The question was if we should favor `p1` over `p2`?

Your examples seems incomplete? In any case, I am suggesting to not look at constraints. Instead, we may have

```python
def f(x: A[T]) -> B[T]: ...
def f(x: A[int]) -> B[int]: ...
```
both in the same pipeline.

SimonHeybrock:
> > I am not sure this is beneficial? It would mean that providers defined in a package can never be used outside the strict scope defined by the package. Say my package defines a generic provider, but I need to add another case, or branch to a workflow. If we enforce constraints this will become impossible.
> 
> In that case, there should be no constraints at all from the first place. If you want to keep them flexible, `bound` should be enough.
> 
> I was thinking about something like:
> 
> ```python
> FlexibleT = TypeVar("FlexibleT", bound=sc.Variable)  # Allows any children of sc.Variable
> StrictT = TypeVar("StrictT", S1, S2)  # Allows only S1 or S2.
> ```

I think this probably makes sense, can we move the discussion to a new issue? I think it is necessary to consider this outside the context of this discussion, since it affect a wider scope of applications.

jokasimr:
Closing because suggestion has been converted to separate issues.

PR #78, opened 2023-12-01
Html repr for pipeline

### LINKED ISSUE #70: Add `_repr_html_` for `TaskGraph`

jokasimr:
Fixes #70

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -26,6 +26,7 @@ from typing import (
 
 from sciline.task_graph import TaskGraph
 
+from .display import ProviderDisplayData, ProviderKind, pipeline_html_repr
 from .domain import Scope, ScopeTwoParams
 from .handler import (
     ErrorHandler,
@@ -37,6 +38,7 @@ from .param_table import ParamTable
 from .scheduler import Scheduler
 from .series import Series
 from .typing import Graph, Item, Key, Label, Provider, get_optional, get_union
+from .utils import groupby, qualname
 
 T = TypeVar('T')
 KeyType = TypeVar('KeyType')
@@ -136,6 +138,14 @@ def _is_multiple_keys(
 
 def provide_none() -> None:
     return None
+
+
+def _kind_of_provider(p: Callable[..., Any]) -> ProviderKind:
+    if qualname(p) == f'{qualname(Pipeline.__setitem__)}.<locals>.<lambda>':
+        return 'parameter'
+    if qualname(p) == f'{qualname(Pipeline.set_param_table)}.<locals>.<lambda>':
+        return 'table'
+    return 'function'
 
 
 class ReplicatorBase(Generic[IndexType]):
@@ -808,3 +818,42 @@ class Pipeline:
 
     def __copy__(self) -> Pipeline:
         return self.copy()
+
+    def _repr_html_(self) -> str:
+        providers_without_parameters = (
+            (p, tuple(), v) for p, v in self._providers.items()
+        )  # type: ignore[var-annotated]
+        providers_with_parameters = (
+            (p, c, v)
+            for p in self._subproviders
+            for c, v in self._subproviders[p].items()
+        )
+        providers = groupby(
+            lambda p: p.kind,
+            (
+                ProviderDisplayData(
+                    origin,
+                    args,
+                    k := _kind_of_provider(value),
+                    value() if k != 'function' else value,
+                )
+                for origin, args, value in chain(
+                    providers_with_parameters, providers_without_parameters
+                )
+            ),
+        )
+        param_tables_by_name = groupby(
+            lambda p: p.origin.label[0].tp, providers['table']
+        )
+        param_tables = []
+        for table_name, table_cells in param_tables_by_name.items():
+            columns = groupby(lambda p: p.origin.tp, table_cells)
+            index = [p.origin.label[0].index for p in next(iter(columns.values()))]
+            param_tables.append(
+                ParamTable(
+                    table_name,
+                    {k: [v.value for v in col] for k, col in columns.items()},
+                    index=index,
+                )
+            )
+        return pipeline_html_repr(providers, param_tables)
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -821,12 +821,12 @@ class Pipeline:
 
     def _repr_html_(self) -> str:
         providers_without_parameters = (
-            (p, tuple(), v) for p, v in self._providers.items()
+            (origin, tuple(), value) for origin, value in self._providers.items()
         )  # type: ignore[var-annotated]
         providers_with_parameters = (
-            (p, c, v)
-            for p in self._subproviders
-            for c, v in self._subproviders[p].items()
+            (origin, args, value)
+            for origin in self._subproviders
+            for args, value in self._subproviders[origin].items()
         )
         providers = groupby(
             lambda p: p.kind,
@@ -834,8 +834,8 @@ class Pipeline:
                 ProviderDisplayData(
                     origin,
                     args,
-                    k := _kind_of_provider(value),
-                    value() if k != 'function' else value,
+                    kind := _kind_of_provider(value),
+                    value() if kind != 'function' else value,
                 )
                 for origin, args, value in chain(
                     providers_with_parameters, providers_without_parameters
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -842,18 +842,4 @@ class Pipeline:
                 )
             ),
         )
-        param_tables_by_name = groupby(
-            lambda p: p.origin.label[0].tp, providers['table']
-        )
-        param_tables = []
-        for table_name, table_cells in param_tables_by_name.items():
-            columns = groupby(lambda p: p.origin.tp, table_cells)
-            index = [p.origin.label[0].index for p in next(iter(columns.values()))]
-            param_tables.append(
-                ParamTable(
-                    table_name,
-                    {k: [v.value for v in col] for k, col in columns.items()},
-                    index=index,
-                )
-            )
-        return pipeline_html_repr(providers, param_tables)
+        return pipeline_html_repr(providers)
--- a/tests/pipeline_test.py
+++ b/tests/pipeline_test.py
@@ -1198,6 +1198,7 @@ def test_pipeline_with_generics_setitem_on_copy_does_not_affect_original() -> No
     assert a.compute(Result) == 29
     assert b.compute(Result) == 53
 
+
 def test_html_repr() -> None:
     pipeline = sl.Pipeline([make_int], params={float: 5.0})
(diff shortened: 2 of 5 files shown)


src/sciline/pipeline.py
```diff
@@ -750,3 +764,52 @@ def bind_and_call(
         if not return_tuple:
             return results[0]
         return results
+
+    def _repr_html_(self) -> str:
```
SimonHeybrock, line 768: Should we adapt some of the docs notebooks to make use of the repr?

jokasimr:
Example
Screenshot from 2023-12-08 16-58-19

jokasimr:
This is what is looks like for the SANS2D workflow in esssans:
Screenshot from 2023-12-08 17-05-44

nvaytet:
Wouldn't the output of `visualize` be a better repr? What was the reason we wanted to have a table?

jokasimr:
I think the motives for having a html repr for the pipeline were:

 * having a more compact description
 * listing information not avaliable in the graphviz, values of parameters etc
 * to be able to display a pipeline without selecting a particular result
 * for very large pipelines inspecting the graph can become cumbersome
 * debugging incomplete pipelines with missing parts
 
Are there other?

jokasimr:
> We also cannot use so much horizontal space, we should show params and providers in a single table, I think?

Sure we can do that but then we will use more vertical space. What's the problem with using the horizontal space?

jokasimr:
New version, docstring removed, added name and source of provider function instead
Screenshot from 2023-12-11 11-40-48

SimonHeybrock:
> Wouldn't the output of `visualize` be a better repr? What was the reason we wanted to have a table?

I think it serves two very different purposes. Compare, e.g., also Dask's `_repr_html_` vs. visualize.

SimonHeybrock:
> > We also cannot use so much horizontal space, we should show params and providers in a single table, I think?
> 
> Sure we can do that but then we will use more vertical space. What's the problem with using the horizontal space?

For example that is does not fit into the docs.

SimonHeybrock:
> > We also cannot use so much horizontal space, we should show params and providers in a single table, I think?
> 
> Sure we can do that but then we will use more vertical space. What's the problem with using the horizontal space?

I am not even sure we want to split providers from params: In different configs of a pipeline we might get the same type from either a provider or a param. Have you considering having everything in a single table, alphabetically?

jokasimr:
Here's a new version.
The `satisfied` column now displays all type var combinations that are possible (based on the constraints of the type var) and highlights the ones that are satisfied.
The `value` column uses the scipp `._repr_html_()` method.
Screenshot from 2023-12-13 16-56-54
Screenshot from 2023-12-13 16-57-28
Screenshot from 2023-12-13 16-57-28

SimonHeybrock:
> The `satisfied` column now displays all type var combinations that are possible (based on the constraints of the type var) and highlights the ones that are satisfied.
> 
> It uses quite a lot of space, but on the other hand it is quite useful to be able to tell quickly what providers exist and are satisfied. One option is to make the list of concrete providers collapsed by default.

I think I'd keep things simple for now and simply remove the `satisfied` column.

src/sciline/pipeline.py
```diff
@@ -750,3 +776,125 @@ def bind_and_call(
...
+                <th scope="col">Name</th>
+                <th scope="col">Source</th>
+                <th scope="col">Value</th>
+              </tr>
+            </thead>
+            <tbody>
+            {chr(10).join((f"""
+              <tr>
+                <th scope="row">{pretty_provider_name(p)}</th>
+                <td scope="row">{provider_source(p)}</th>
```
nvaytet, line 873: I'm getting some strange values for the paths in the source dropdowns:
Screenshot at 2023-12-15 11-25-51

src/sciline/pipeline.py
```diff
@@ -750,3 +776,125 @@ def bind_and_call(
...
+              <tr>
+                <th scope="col">Name</th>
+                <th scope="col">Source</th>
+                <th scope="col">Value</th>
+              </tr>
+            </thead>
+            <tbody>
+            {chr(10).join((f"""
+              <tr>
+                <th scope="row">{pretty_provider_name(p)}</th>
```
nvaytet, line 872: Do we really need to have the names in bold?

src/sciline/pipeline.py
YooSunYoung: Can we have information of `package/module/class` it's coming from?
Instead of the source file and number of line.

jokasimr: Yes package/module/function_name makes more sense than file and line number. I'll see if I can fix that.

src/sciline/pipeline.py
jokasimr: No not really

src/sciline/pipeline.py
```diff
@@ -750,3 +776,124 @@ def bind_and_call(
...
+
+        def provider_value(p: dict[str, Any]) -> str:
+            if p['kind'] != 'parameter':
+                return ''
+            if hasattr(p['value'], '_repr_html_'):
+                return f'{p["value"]._repr_html_()}'
+            return f'{p["value"]}'
+
+        return f'''
+        <div style="display: flex; column-gap: 1em; align-items: flex-start;">
```
YooSunYoung, line 849: Should we maybe use template instead of f-string here...?
Then it'll be much easier to update/fix them later.
Llike for the data group html wrapper: https://github.com/scipp/scipp/blob/main/src/scipp/visualization/formatting_datagroup_html.py

Or maybe later if we want to add more columns...!

jokasimr: I agree the code needs a clean up. I'll collect the inline styles and remove excessive duplication and nested f-strings etc. I'm not sure that the refactor will involve templates, I'll look into it but it might be overkill here.

SimonHeybrock: Maybe you can also move this to a different file to not grow `pipeline.py` too much?

src/sciline/pipeline.py
```diff
@@ -750,3 +760,42 @@ def bind_and_call(
...
+                    origin,
+                    args,
+                    k := _kind_of_provider(value),
+                    value() if k != 'function' else value,
+                )
+                for origin, args, value in chain(
+                    providers_with_parameters, providers_without_parameters
+                )
+            ),
+        )
```
SimonHeybrock, line 786: Will this make one entry in the repr **per row** (and column) of a param table? Or am I misreading this?

jokasimr: You're misreading it. Take a look at the examples in the screenshots I've posted in this PR. Parameter Tables use their own built in html repr.

However, it is quite messy to reconstruct the parameter tables from the `_provider` and `_subprovider` dicts, so it would be good if we could do that in a better way.

SimonHeybrock: I thought we had decided to combine everything into a single table? I had thought those screenshots were thus outdated.

jokasimr: Okay I see. The parameters were merged with the providers in a single table. But the parameter tables are still displayed separately. Do you think it'd be better to merge them in the larger table as well? I think that it might be more messy. The parameter tables are self contained and have a html repr already so I thought it sensible to use that and display them separately.

SimonHeybrock: Imho a user can display the param table themselves if they want to. The HTML repr *of the pipeline should* give info *about the pipeline* (i.e., mainly the structure), and as such I would consider each param table "column" as a single entity. That is, simply display a single summary line in the repr, instead of all the values. After all there might be thousands of rows in a param table?

jokasimr: That makes a lot of sense. I'll do something like that.

jokasimr: Screenshot from 2023-12-20 14-23-42

jokasimr: I decided to put values of parameters in the table, but below a `<details>` tag. This way they don't take up space when they are not expanded, but can be quickly expanded and displayed using their own `_repr_html_`.

jokasimr: Screenshot from 2023-12-20 14-27-57

SimonHeybrock: Can we swap the "Source" and "Value" columns

SimonHeybrock: As I said before, I would be worried about size and lag when using this with large param tables. I think we should not include full nested HTML reprs of each value. Listing the size should be a good start.

It is generally always easier to make things more complicated later, than to remove it (because then some users who learned to love and rely on a feature will become upset).

jokasimr: I agree it might be a problem, but I think it is a unusual problem and one that is difficult to protect against in general. Displaying the html repr of any DataArray might lag if it is large enough. I also think being able to view the values of all parameters in the workflow like this is quite useful.

We talked before about displaying a reduced view of variables like is done in datagroup. The issue with that approach in this case is that we don't want the sciline code to contain lots of logic about how to display scipp datatypes.

I don't know what you mean by listing the size, values of parameters can have any type, they have no concept of size.

SimonHeybrock: > I'm not sure about what you mean by listing the size, values of parameters can have any type and they don't have a common concept of size.

I meant the number of rows.

jokasimr: Yes it switched a bit back and forth and I understand it got confusing.

When we talked at my desk the other day we decided to only display the length of parameter tables. That's what's done now.

Now I thought we were talking about *parameters* that are not in tables. Just regular input values like `QBins` or `Filename[Sample`. Those are currently displayed below a details tag in case they have a html repr. Otherwise they're just stringified.

> Shouldn't the param-table info go into the source column, not the value column? Something like ParamTable(int, length=1000) maybe?

That sounds like a good idea. Maybe the source column should change name to "info" or something like that instead.

SimonHeybrock:
Can you post some screenshots of the latest implementation?

Also, you have some merge conflicts to address.

jokasimr:
Here are some screenshots:
One with collapsed fields and second with uncollapsed.
The `value` field is collapsed or not based on a cutoff for the string length.
Screenshot from 2024-01-10 11-34-34
Screenshot from 2024-01-10 11-35-27

SimonHeybrock:
Can you also show one with a param table?

tests/pipeline_test.py
```diff
@@ -1197,3 +1197,8 @@ def process(
     b[RawData[Sample]] = 7
     assert a.compute(Result) == 29
     assert b.compute(Result) == 53
+
+
+def test_html_repr() -> None:
+    pipeline = sl.Pipeline([make_int], params={float: 5.0})
```
SimonHeybrock, line 1203: Can we add generic providers and a param table for this (or separate) test, just to make sure this doesn't break the repr?

jokasimr:
Here is a screenshot for a case with parameter tables:
Screenshot from 2024-01-10 11-49-36

SimonHeybrock (approved):

PR #80, opened 2023-12-05
Allow visualization of graphs with missing requirements

### LINKED ISSUE #79: Visualize graphs with missing dependencies
Currently, we cannot visualize a graph unless the are providers or parameters for all dependencies. It would be very useful if we could visualize a graph even in case of missing requirements, e.g., marking the missing ones in red.

Implementation wise, I think this could be achieved by passing a *handler* to the `get` or `visualize` methods of `Pipeline`. The default handler would raise `UnsatisfiedRequirement`, i.e., what the current implementation does. A custom handler could return a dummy, which could be used by, e.g., the visualization code to draw a helpful graph for debugging, or for developing a new workflow without setting all params.

SimonHeybrock:
Fixes #79.

This avoids wasting a lot of time when working with scientists to draft a workflow. Previously I had to add a ton of dummy providers. I think this may also be helpful later to highlight which parameters still need to be set.

Currently this only works for the non-param-table case, we can add support for that later.

Example:
image

### DIFF
--- a/src/sciline/handler.py
+++ b/src/sciline/handler.py
@@ -1,0 +1,52 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
+from __future__ import annotations
+
+from typing import Callable, Protocol, Type, TypeVar, Union
+
+from .typing import Item
+
+T = TypeVar('T')
+
+
+class UnsatisfiedRequirement(Exception):
+    """Raised when a type cannot be provided."""
+
+
+class Handler(Protocol):
+    """Error handling protocol for pipelines."""
+
+    def handle_unsatisfied_requirement(self, tp: Union[Type[T], Item[T]]):
+        ...
+
+
+class HandleAsBuildTimeException(Handler):
+    """
+    Error handler used by default.
+
+    This will raise an exception when building the graph, which is helpful for
+    ensuring that errors are caught early, before starting costly computation.
+    """
+
+    def handle_unsatisfied_requirement(self, tp: Union[Type[T], Item[T]]):
+        """Raise an exception when a type cannot be provided."""
+        raise UnsatisfiedRequirement('No provider found for type', tp)
+
+
+class HandleAsComputeTimeException(Handler):
+    """
+    Error handler used for Pipeline.visualize().
+
+    This avoids raising exceptions when building the graph, which would prevent
+    visualization. This is helpful when visualizing a graph that is not yet complete.
+    """
+
+    def handle_unsatisfied_requirement(
+        self, tp: Union[Type[T], Item[T]]
+    ) -> Callable[[], T]:
+        """Return a function that raises an exception when called."""
+
+        def unsatisfied_sentinel() -> None:
+            raise UnsatisfiedRequirement('No provider found for type', tp)
+
+        return unsatisfied_sentinel
--- a/src/sciline/handler.py
+++ b/src/sciline/handler.py
@@ -13,7 +13,7 @@ class UnsatisfiedRequirement(Exception):
     """Raised when a type cannot be provided."""
 
 
-class Handler(Protocol):
+class ErrorHandler(Protocol):
     """Error handling protocol for pipelines."""
 
     def handle_unsatisfied_requirement(
@@ -22,7 +22,7 @@ class Handler(Protocol):
         ...
 
 
-class HandleAsBuildTimeException(Handler):
+class HandleAsBuildTimeException(ErrorHandler):
     """
     Error handler used by default.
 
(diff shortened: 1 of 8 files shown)


src/sciline/handler.py
```diff
@@ -0,0 +1,54 @@
...
+from .typing import Item
+
+T = TypeVar('T')
+
+
+class UnsatisfiedRequirement(Exception):
+    """Raised when a type cannot be provided."""
+
+
+class Handler(Protocol):
```
jl-wynen, line 16:
```suggestion
class ErrorHandler(Protocol):
```
And rename the file.

Or pick a different name. But 'handler' is too generic.

src/sciline/pipeline.py
```diff
@@ -511,7 +527,9 @@ def build(
             if (optional_arg := get_optional(tp)) is not None:
                 try:
                     optional_subgraph = self.build(
-                        optional_arg, search_param_tables=search_param_tables
+                        optional_arg,
+                        search_param_tables=search_param_tables,
+                        handler=HandleAsBuildTimeException(),
```
jl-wynen, line 532: Why does this use a build time error instead of forwarding the handler?

src/sciline/pipeline.py
```diff
@@ -477,10 +488,15 @@ def _get_provider(
...
-        raise UnsatisfiedRequirement("No provider found for type", tp)
+
+        return handler.handle_unsatisfied_requirement(tp), {}
 
     def build(
-        self, tp: Union[Type[T], Item[T]], /, search_param_tables: bool = False
+        self,
+        tp: Union[Type[T], Item[T]],
+        /,
+        handler: Handler,
```
jl-wynen, line 498: Should this be akeyword-only arg?

jl-wynen, line 498: Should this have a default value?

src/sciline/pipeline.py
```diff
@@ -593,7 +611,9 @@ def _build_series(self, tp: Type[Series[KeyType, ValueType]]) -> Graph:
         value_type: Type[ValueType]
         index_name, value_type = get_args(tp)
 
-        subgraph = self.build(value_type, search_param_tables=True)
+        subgraph = self.build(
+            value_type, search_param_tables=True, handler=HandleAsBuildTimeException()
```
jl-wynen, line 615: See above for forwarding handler.

src/sciline/pipeline.py
```diff
@@ -296,6 +298,14 @@ def __call__(self, *vals: ValueType) -> Series[KeyType, ValueType]:
         return Series(self.row_dim, dict(zip(self.labels, vals)))
 
 
+class unsatisfied_sentinel:
```
jl-wynen, line 301: Unused?

src/sciline/visualize.py
```diff
@@ -104,13 +105,24 @@ def get_subgraph_name(name: str) -> str:
...
-            ret.name, ret.name, shape='box3d' if ret.collapsed else 'rectangle'
-        )
+        unsatisfied = f'{_qualname(HandleAsComputeTimeException.handle_unsatisfied_requirement)}.<locals>.unsatisfied_sentinel'  # noqa: E501
+        if p_name == unsatisfied:
+            subgraph.node(
+                ret.name,
+                ret.name,
+                shape='box3d' if ret.collapsed else 'rectangle',
+                color='red',
+                fontcolor='red',  # Set text color to red
```
jl-wynen, line 115: Can you add another visual hint in addition to colour? E.g. make the border dashed.

src/sciline/pipeline.py
SimonHeybrock: This is handling an "optional" branch. I thought if a requirement is optional we should prune the entire branch, as before?

src/sciline/pipeline.py
SimonHeybrock: See PR message: This is not supported for param tables yet.

src/sciline/pipeline.py
SimonHeybrock: Not really. Users should be using `get`.

src/sciline/pipeline.py
jl-wynen: Makes sense. But we could also show optional requirements in the viz. But that should be a separate PR. And it's not clear to me whether we'd want to show only the immediate requirement or all transitive requirements, too.

src/sciline/pipeline.py
jl-wynen: Then, should `build` not be public?

SimonHeybrock: I think there is some value in having it. And the docstring is explicit about it, and refers to `get`.

src/sciline/visualize.py
SimonHeybrock: Done

jl-wynen (approved):

src/sciline/pipeline.py
SimonHeybrock: Currently it shows the immediate requirements.

Issue #43, opened 2023-08-31
Detect unused parameters


SimonHeybrock:
Having the ability to detect parameters that are set but unused would greatly simplify debugging, but also allow for doing "validated" computation, which could raises if a parameter is not used.

As we may want to compute only an intermediate result (which would typically not depend on all parameters) this should not be enforced or enabled by default. That is, provide another helper, or argument.

SimonHeybrock:
Might also want to check for unused providers. This should probably be an independent flag.

YooSunYoung:
Maybe we could just extract a graph from the pipeline 
and use other packages to detect missing parameters or providers,
instead of checking them manually (with `if` or `for` statement).

Something like 
```python
import networkx as nx
G = nx.Graph()
G.add_nodes_from([1, 2, 3, 4, 5])
G.add_edge(1, 2)
G.add_edge(1, 3)
list(nx.connected_components(G))

# [{1, 2, 3}, {4}, {5}]
```

And then we can easily find all of the disconnected components and do something about them, finding unused params or providers or missing providers etc....

Is there an easy way to retrieve edges and nodes from ``graphviz.Digraph``.....?

jl-wynen:
> Is there an easy way to retrieve edges and nodes from graphviz.Digraph.....?

If we implement several diagnostic tools, we should have a way of directly constructing a graph (networkx directly or a custom representation). Going for networkx directly might be the best choice because it gives us access to a lot of tools like @YooSunYoung showed.

SimonHeybrock:
I may furthermore be useful to detect unused providers.

Issue #74, opened 2023-11-20
More helpful error messages when pipeline fails to build or compute?


nvaytet:
When a pipeline fails to build or compute, the error message gives you the type it failed with.
Not sure how feasible it is, but it would be even more helpful if it printed the name of the provider it failed in, as the same types can be used by multiple providers.

This would make debugging pipelines a lot faster.

SimonHeybrock:
Use Python 3.10 or higher, so `NewType` is implemented as a class instead of a function. You will then get proper errors.

As Python 3.8 and 3.9 are/will be dropped soon I will close this.

jokasimr:
It seems feasible. We could add both `tp` and the provider that required `tp` to the stack [here](https://github.com/scipp/sciline/blob/2547d90fd85c1b80d6a2521b205d80bd97fecd65/src/sciline/pipeline.py#L549) and then make the provider available when the error is raised, either by passing it as an extra argument to `_get_provider` or by catching the exceptions raised by `_get_provider` and editing the error messages to add the name of the provider.

SimonHeybrock:
Conceptually this is not really (or easily) possible: There may be multiple providers that require a certain type. The way the graph is build (non-recursively) we do not necessary know which provider "required" something. We could add tracking via list of dependees. See
https://github.com/scipp/sciline/blob/2547d90fd85c1b80d6a2521b205d80bd97fecd65/src/sciline/pipeline.py#L546-L549

Would it make more sense to generalize other available diagnostic tools? See also my comment https://github.com/scipp/sciline/pull/86#discussion_r1429595333.

Basically, #74 and #86 are just two statements about "invalid" graphs: Graph is disconnected, or graph is incomplete. By adding "piecemeal" solutions we risk making the code hard to maintain. I suggest we take a step back and think about the bigger picture before moving forward with any of these.

PR #86, opened 2023-12-14
feat: add check to determine if graph uses all avaliable parameters

jokasimr:
Fixes [#43](https://github.com/scipp/sciline/issues/43).

I wasn't sure if this should be added to the task graph rather than to the pipeline. What do you think?

I think there is also an opportunity here to extract the functionality of retrieving all providers to a helper method, it's also used in the `_repr_html_` for example, but I didn't want to complicate things for now.

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -656,15 +656,15 @@ class Pipeline:
         return graph
 
     @overload
-    def compute(self, tp: Type[T], check: bool) -> T:
+    def compute(self, tp: Type[T], check: bool = ...) -> T:
         ...
 
     @overload
-    def compute(self, tp: Iterable[Type[T]], check: bool) -> Dict[Type[T], T]:
+    def compute(self, tp: Iterable[Type[T]], check: bool = ...) -> Dict[Type[T], T]:
         ...
 
     @overload
-    def compute(self, tp: Item[T], check: bool) -> T:
+    def compute(self, tp: Item[T], check: bool = ...) -> T:
         ...
 
     def compute(self, tp: type | Iterable[type] | Item[T], check: bool = False) -> Any:
@@ -705,7 +705,7 @@ class Pipeline:
         *,
         scheduler: Optional[Scheduler] = None,
         handler: Optional[ErrorHandler] = None,
-        check=False,
+        check: bool = False,
     ) -> TaskGraph:
         """
         Return a TaskGraph for the given keys.
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -72,6 +72,20 @@ def _is_compatible_type_tuple(
     return True
 
 
+def _qualname(obj: Any) -> Any:
+    return (
+        obj.__qualname__ if hasattr(obj, '__qualname__') else obj.__class__.__qualname__
+    )
+
+
+def _kind_of_provider(p: Callable[..., Any]) -> str:
+    if _qualname(p) == f'{_qualname(Pipeline.__setitem__)}.<locals>.<lambda>':
+        return 'parameter'
+    if _qualname(p) == f'{_qualname(Pipeline.set_param_table)}.<locals>.<lambda>':
+        return 'table'
+    return 'function'
+
+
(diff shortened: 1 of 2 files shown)


src/sciline/pipeline.py
```diff
@@ -72,6 +72,20 @@ def _is_compatible_type_tuple(
...
+def _qualname(obj: Any) -> Any:
+    return (
+        obj.__qualname__ if hasattr(obj, '__qualname__') else obj.__class__.__qualname__
+    )
+
+
+def _kind_of_provider(p: Callable[..., Any]) -> str:
+    if _qualname(p) == f'{_qualname(Pipeline.__setitem__)}.<locals>.<lambda>':
+        return 'parameter'
+    if _qualname(p) == f'{_qualname(Pipeline.set_param_table)}.<locals>.<lambda>':
```
YooSunYoung, line 84: Will it be better to make a dummy ``Pipeline`` with a parameter and a table
and retrieve their names, and use them to check these instead?

Or if you add the passing case unit test I mentioned, it might be fine like this.

tests/pipeline_test.py
```diff
@@ -917,3 +917,12 @@ def func(x: int) -> float:
...
         graph.compute()
+
+
+def test_pipeline_with_unused_parameter_raises_if_check() -> None:
+    Unused = NewType('Unused', str)
+    pipeline = sl.Pipeline([int_to_float, make_int])
+    pipeline[Unused] = 'Should raise'
+    assert pipeline.compute(float) == 1.5
+    with pytest.raises(RuntimeError):
+        assert pipeline.compute(float, check=True) == 1.5
```
YooSunYoung, line 928: Can you also add a test that pass,
even if the ``compute`` passes with the ``check=True``?

src/sciline/pipeline.py
```diff
@@ -710,6 +725,8 @@ def get(
             During development and debugging it can be helpful to use a handler that
             raises an exception only when the graph is computed. This can be achieved
             by passing :py:class:`HandleAsComputeTimeException` as the handler.
+        check:
+            Checks if the constructed graph uses all available parameters.
```
YooSunYoung, line 729: In #43, detecting missing providers was also mentioned, so maybe we need a more specific name for unused parameters check here.

src/sciline/pipeline.py
jokasimr: I don't quite understand what you mean. Do you mean to compute the expected `_qualname`s instead of hardcoding something like `f'{_qualname(Pipeline.set_param_table)}.<locals>.<lambda>'`?

YooSunYoung: No sth like.. 
```python 
_DUMMY_PIPELINE = sl.Pipeline(params={int: 0})
_PARAMETER_PROVIDER_NAME = _DUMMY_PIPELINE._get_provider(int)[0].__qualname__
```

src/sciline/pipeline.py
SimonHeybrock: Yes, we need a more meaningful name.

src/sciline/pipeline.py
```diff
@@ -719,6 +736,23 @@ def get(
...
+                    for c, v in self._subproviders[p].items()
+                ),
+                ((p, v) for p, v in self._providers.items()),
+            )
+
+            for p, v in all_providers:
+                if p not in graph and _kind_of_provider(v) == 'parameter':
+                    raise RuntimeError(
+                        'Graph was expected to use all available parameters.'
+                    )
```
SimonHeybrock, line 754: If we take a step back, I am not sure this is sufficiently capturing the problem. In general, I think what we want to do is check whether a graph is connected, I suppose? Raising errors about individual disconnected pieces may not be meaningful?

Instead of the `check` flag and handling things via exceptions, would it be better to return some list (or other data structure) of disconnected pieces of the graph?

SimonHeybrock:
See https://github.com/scipp/sciline/issues/74#issuecomment-1860373045, I suggest to put this on hold so we can have a think.

PR #88, opened 2023-12-18
Drop Python 3.8 support

### LINKED ISSUE #60: Drop Python 3.8 (and 3.9?) support
We should be good to do this now, since Mantid is moving forward.

SimonHeybrock:
Fixes #60.


jl-wynen (approved):
It's ok now. But can we in future separate changes to python or other versions from copier updates? This would make it easier to track down and revert problems.

SimonHeybrock:
> It's ok now. But can we in future separate changes to python or other versions from copier updates? This would make it easier to track down and revert problems.

Hmm, you mean rerun copier, but if it the old ref?

jl-wynen:
No. Do a copier update first, commit, and then run it again and change the version.

PR #89, opened 2023-12-18
Support Union requirements, satiesfied by unique match among the options

### LINKED ISSUE #75: Alternative types in parameters
## Problem

For some parameters, it makes sense to be flexible about what types are allowed. For example:

- Numeric parameters are often `float`. But in Python, passing an `int` instead would be expected to work the same.
- File paths are commonly encoded as either `str` or `pathlib.Path`.

But `Pipeline` is strict about the types. E.g., 
```python
params = {
    float: 3
}
sciline.Pipeline((), params=params)
```
raises `TypeError: Key <class 'float'> incompatible to value 3 of type <class 'int'>`

### Convert to the declared type

Change https://github.com/scipp/sciline/blob/f04f65e5f9fec94987de5713b823912af51dfad8/src/sciline/pipeline.py#L384-L386 to not raise an exception but attempt a conversion instead. I.e., something along the lines of (with better error reporting)
```python
self._set_provider(key, lambda: expected(param))
```
Or (for earlier error detection but more expensive pipeline construction)
```python
converted = expected(param)
self._set_provider(key, lambda: converted)
```

This would make the pipeline almost as flexible as Python in terms of typing. It wouldn't quite be duck-typing as it requires types that are convertible to `expected` but it might be close enough.

Potential downsides are that these conversions can be expensive and that this can hide mistakes. But both are typical behaviour in Python programs.

### Do not check the provided type

Simply drop the type-checking code shown above. This would correspond to regular duck-typing. If the provided value is of an unsupported type, this will lead to a runtime error in the providers that consume it.

So this would be maximally flexible but also shift right the error detection.

## Workaround for numbers

In the case of int/float, parameters can instead be declared like this:
```python
P1 = NewType('P1', Real)
P2 = NewType('P2', Real)
params = {
    P1: 3,   # works
    P2: 3.0, # works, too
}
sciline.Pipeline((), params=params)
```
But this only works where a common base type is available. And the pipeline author needs to design the pipeline for it.

### Use Union

It is tempting to use
```python
P = NewType('P', Union[str, Path])
params = {
    P: 'file'
}
sciline.Pipeline((), params=params)
```
but Sciline doesn't like unions as provider types. The above raises 
```
  File ".../site-packages/sciline/pipeline.py", line 356, in __setitem__
    elif issubclass(origin, (Scope, ScopeTwoParams)):
TypeError: issubclass() arg 1 must be a class
```

Doing this with a scoped parameter:
```python
class P(sciline.Scope[RunType, Union[str, Path]], Union[str, Path]): ...
```
raises
```
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

### Use TypeVar

Defining the parameter as a type var:
```python
P = TypeVar('P', int, float)
params = {
    P: 3
}
sciline.Pipeline((), params=params)
```
raises
```
  File ".../site-packages/sciline/pipeline.py", line 367, in __setitem__
    if not isinstance(param, expected):
TypeError: isinstance() arg 2 must be a type or tuple of types
```

SimonHeybrock:
Fixes #75.

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -36,7 +36,7 @@ from .handler import (
 from .param_table import ParamTable
 from .scheduler import Scheduler
 from .series import Series
-from .typing import Graph, Item, Key, Label, Provider, get_optional
+from .typing import Graph, Item, Key, Label, Provider, get_optional, get_union
 
 T = TypeVar('T')
 KeyType = TypeVar('KeyType')
@@ -536,7 +536,23 @@ class Pipeline:
                     graph.update(optional_subgraph)
                 continue
             provider: Callable[..., T]
-            provider, bound = self._get_provider(tp, handler=handler)
+            if (union_args := get_union(tp)) is None:
+                provider, bound = self._get_provider(tp, handler=handler)
+            else:
+                # When requiring a Union we look for a unique match among the union args
+                matches = 0
+                for arg in union_args:
+                    try:
+                        provider, bound = self._get_provider(arg, handler=None)
+                        matches += 1
+                    except UnsatisfiedRequirement:
+                        continue
+                if matches == 0:
+                    handler.handle_unsatisfied_requirement(tp)
+                if matches > 1:
+                    raise AmbiguousProvider(
+                        f"Multiple providers found for Union type {tp}."
+                    )
             tps = get_type_hints(provider)
             args = tuple(
                 _bind_free_typevars(t, bound=bound)
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -493,8 +493,7 @@ class Pipeline:
         """Get a unique provider for a potential Union type."""
         if (union_args := get_union(tp)) is None:
             return self._get_provider(tp, handler=handler)
-        provider: Callable[..., T]
-        matching_types: list[Union[Type[T], Item[T]]] = []
+        matching_types = []
         for option in union_args:
             try:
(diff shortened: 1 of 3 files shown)


YooSunYoung:
How will it look like in the graph or html wrapper...?

YooSunYoung:
Should it be also allowed to have ``Union`` as a return type...?

For example, if here is ``provider_a`` that returns ``Union[int, float]``
and ``provider_b`` that needs ``Union[int, float, str]`` for an argument ``a``,
it'll not find the ``provider_a`` as a provider of ``a``,
But if there is ``provider_c`` that returns ``Union[int, float, str]`` it should be able to use ``provider_c`` as a provider of ``a``.

Maybe I'm overthinking about it...

SimonHeybrock:
> How will it look like in the graph or html wrapper...?

It will show the match.

SimonHeybrock:
> Should it be also allowed to have `Union` as a return type...?
> 
> For example, if here is `provider_a` that returns `Union[int, float]` and `provider_b` that needs `Union[int, float, str]` for an argument `a`, it'll not find the `provider_a` as a provider of `a`, But if there is `provider_c` that returns `Union[int, float, str]` it should be able to use `provider_c` as a provider of `a`.
> 
> Maybe I'm overthinking about it...

I feel this may be a bad idea. I think we should not support it for now.

YooSunYoung:
I think we also need a short description of how ``Union`` annotation works in the documentation.

YooSunYoung (approved):

PR #90, opened 2023-12-19
Support replacing params and providers

### LINKED ISSUE #57: Support replacing providers based on their key, i.e., their return value
After using Sciline for a little, it has become evident that (as anticipated) we need a more convenient way of replacing providers. There are two options:

- Remove the checks in `__setitem__`, `set_provider`, ... which prevent replacements. Note that this may be tricky, since we can have parameters, providers, generic providers, as well as param tables. That is, care must be taken (and a lot of tests be written).
- Simpler options: Provide dict-like helpers to store providers *prior* to pipeline creation. Users can currently word around this by writing a simple `as_dict` function that parses the return type hints and uses them as keys.

SimonHeybrock:
Fixes #57

### DIFF
--- a/tests/pipeline_test.py
+++ b/tests/pipeline_test.py
@@ -577,11 +577,114 @@ def test_setitem_raises_TypeError_if_instance_does_not_match_key() -> None:
         pl[B[int]] = 1.0
 
 
-def test_setitem_raises_if_key_exists() -> None:
+def test_setitem_can_replace_param_with_param() -> None:
     pl = sl.Pipeline()
     pl[int] = 1
-    with pytest.raises(ValueError):
-        pl[int] = 2
+    pl[int] = 2
+    assert pl.compute(int) == 2
+
+
+def test_insert_can_replace_param_with_provider() -> None:
+    def func() -> int:
+        return 2
+
+    pl = sl.Pipeline()
+    pl[int] = 1
+    pl.insert(func)
+    assert pl.compute(int) == 2
+
+
+def test_setitem_can_replace_provider_with_param() -> None:
+    def func() -> int:
+        return 2
+
+    pl = sl.Pipeline()
+    pl.insert(func)
+    pl[int] = 1
+    assert pl.compute(int) == 1
+
+
+def test_insert_can_replace_provider_with_provider() -> None:
+    def func1() -> int:
+        return 1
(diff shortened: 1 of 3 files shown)


tests/pipeline_test.py
```diff
@@ -577,11 +577,114 @@ class B(Generic[T]):
...
+
+def test_setitem_can_replace_provider_with_param() -> None:
+    def func() -> int:
+        return 2
+
+    pl = sl.Pipeline()
+    pl.insert(func)
+    pl[int] = 1
+    assert pl.compute(int) == 1
+
```
YooSunYoung, line 605:
```suggestion

def test_setitem_can_replace_instance() -> None:
    pl = sl.Pipeline()
    original_number = 1
    pl[int] = original_number
    assert pl.compute(int) is original_number
    new_number = 1
    pl[int] = new_number
    assert pl.compute(int) == original_number
    assert pl.compute(int) is not original_number
    assert pl.compute(int) is new_number

```

Can we also have a test case that shows instance check like this...?
Just to show that it's passed by lambda function, so the instance is kept not just the value.

SimonHeybrock: `int` is immutable, not sure I see the value of such a test?

tests/pipeline_test.py
```diff
@@ -577,11 +577,114 @@ class B(Generic[T]):
...
+    @dataclass
+    class A(Generic[T]):
+        value: T
+
+    pl = sl.Pipeline()
+    pl[A[T]] = A[T](1)  # type: ignore[valid-type]
+    assert pl.compute(A[int]) == A[int](1)
+    pl[A[T]] = A[T](2)  # type: ignore[valid-type]
+    assert pl.compute(A[int]) == A[int](2)
```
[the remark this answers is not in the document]
SimonHeybrock: This is unrelated to this change, I think? There are already tests for specializations.

tests/pipeline_test.py
YooSunYoung: Yes, but it shows that it doesn't copy the value and return it but keep the instance itself.
In case we want to pass around ``dict``, it might be useful information I thought...?

I just suggested what I wanted to try with the changes.
Feel free to drop them...!

PR #91, opened 2024-01-05
Add pipeline copy method

nvaytet:
This makes it possible to make a copy of an existing pipeline.
One can then, for example modify the parameters of the new pipeline.

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -322,9 +322,6 @@ class Pipeline:
         params:
             Dictionary of concrete values to provide for types.
         """
-        self._copy_providers = list(providers or [])
-        self._copy_params: Dict[type, Any] = {}
-
         self._providers: Dict[Key, Provider] = {}
         self._subproviders: Dict[type, Dict[Tuple[Key | TypeVar, ...], Provider]] = {}
         self._param_tables: Dict[Key, ParamTable] = {}
@@ -347,7 +344,6 @@ class Pipeline:
         if (key := get_type_hints(provider).get('return')) is None:
             raise ValueError(f'Provider {provider} lacks type-hint for return value')
         self._set_provider(key, provider)
-        self._copy_providers.append(provider)
 
     def __setitem__(self, key: Type[T], param: T) -> None:
         """
@@ -391,7 +387,6 @@ class Pipeline:
                 f'Key {key} incompatible to value {param} of type {type(param)}'
             )
         self._set_provider(key, lambda: param)
-        self._copy_params[key] = param
 
     def set_param_table(self, params: ParamTable) -> None:
         """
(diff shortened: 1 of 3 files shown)


tests/pipeline_test.py
jl-wynen, line 1: Missing tests that modifying `a` or `b` after the copy does not affect the other pipeline.

src/sciline/pipeline.py
```diff
@@ -322,6 +322,9 @@ def __init__(
         params:
             Dictionary of concrete values to provide for types.
         """
+        self._copy_providers = list(providers or [])
+        self._copy_params: Dict[type, Any] = {}
```
jl-wynen, line 326: I don't understand why those are needed? Can't you just make copies of `self._providers`, etc.?

nvaytet: No because `self._providers` contains all the original providers, plus a whole bunch of lambda functions that were made from the parameters, which then fail if you try to construct a new pipeline with them.

Also, going through the sub-providers is a bit of a mess, especially with generics. So I thought just saving the constructor params was far easier. But maybe @SimonHeybrock has a better idea?

SimonHeybrock: Why are the lambdas failing?

SimonHeybrock:
Looking at `Pipeline`:
```python
        self._providers: Dict[Key, Provider] = {}
        self._subproviders: Dict[type, Dict[Tuple[Key | TypeVar, ...], Provider]] = {}
        self._param_tables: Dict[Key, ParamTable] = {}
        self._param_name_to_table_key: Dict[Key, Key] = {}
```
Why can't we just copy these four dicts, and copy the testing dicts in `_subproviders`?

src/sciline/pipeline.py
nvaytet: `ValueError: Provider <function Pipeline.__setitem__.<locals>.<lambda> at 0x7f492825f130> lacks type-hint for return value`

nvaytet:
>  and copy the nested dicts in `_subproviders`

Can you have more than one level of nesting?

src/sciline/pipeline.py
SimonHeybrock: The lambda is raising that? That is really odd.

SimonHeybrock:
> > and copy the nested dicts in `_subproviders`
> 
> Can you have more than one level of nesting?

No.

src/sciline/pipeline.py
```diff
@@ -794,3 +794,14 @@ def bind_and_call(
         if not return_tuple:
             return results[0]
         return results
+
+    def copy(self) -> Pipeline:
```
SimonHeybrock, line 798: Also implement `__copy__`?

nvaytet: Sure, can do.

jl-wynen: Do we actually need a custom `copy` implementation? Or would this be enough with the default `__copy__`?
```python
def copy(self) -> Pipeline:
    from copy import copy
    return copy(self)
```

jl-wynen: Right! It's deep as far as the dicts are concerned but not the dict elements.

src/sciline/pipeline.py
nvaytet: Just to be clear, I get the error from the lambdas when I do:
```Py
Pipeline(existing_pipeline._providers.values())
```

nvaytet: Making a copy of the dicts internally works fine.

PR #99, opened 2024-01-11
feat: task graph html repr

### LINKED ISSUE #70: Add `_repr_html_` for `TaskGraph`

jokasimr:
Fixes #70

### DIFF
--- a/src/sciline/task_graph.py
+++ b/src/sciline/task_graph.py
@@ -2,12 +2,60 @@
 # Copyright (c) 2023 Scipp contributors (https://github.com/scipp)
 from __future__ import annotations
 
-from typing import Any, Optional, Tuple, TypeVar, Union
+from html import escape
+from typing import Any, Optional, Sequence, Tuple, TypeVar, Union
 
 from .scheduler import DaskScheduler, NaiveScheduler, Scheduler
 from .typing import Graph, Item
+from .utils import keyname, kind_of_provider
 
 T = TypeVar("T")
+
+
+def _list_items(items: Sequence[str]) -> str:
+    return '\n'.join(
+        (
+            '<ul>',
+            ('\n'.join((f'<li>{escape(it)}</li>' for it in items))),
+            '</ul>',
+        )
+    )
+
+
+def _list_max_n_then_hide(items: Sequence[str], n: int = 5, header: str = '') -> str:
+    def wrap(s: str) -> str:
+        return '\n'.join(
+            (
+                '<div class="task-graph-detail-list">'
+                '<style> .task-graph-detail-list ul { margin-top: 0; } </style>',
+                s,
+                '</div>',
+            )
+        )
+
+    return wrap(
+        '\n'.join(
+            (
+                header,
+                _list_items(items),
+            )
+        )
+        if len(items) <= n
+        else '\n'.join(
+            (
+                '<details>',
+                '<style>',
+                'details[open] .task-graph-summary ul { display: none; }',
+                '</style>',
+                '<summary class="task-graph-summary">',
+                header,
+                _list_items((*items[:n], '...')),
+                '</summary>',
+                _list_items(items),
+                '</details>',
+            )
+        )
+    )
 
 
 class TaskGraph:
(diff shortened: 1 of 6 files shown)


jokasimr:
Here's a suggestion for how the task graph repr could look:
Screenshot from 2024-01-11 17-42-10

jokasimr:
Case with ParamTables:
Screenshot from 2024-01-11 18-05-56

src/sciline/task_graph.py
```diff
@@ -79,3 +109,30 @@ def visualize(
...
+            }
+        )
+        scheduler = escape(self._scheduler.__class__.__name__)
+        return ''.join(
+            (
+                '<p>Task graph computes</p>\n',
+                _list_items(leafs),
+                '<p>based on</p>\n',
+                _list_max_n_then_hide(roots),
+                f'<p>using the scheduler {scheduler}.</p>',
```
SimonHeybrock, line 136: I must say not a fan of the "narrative" formatting.  How about:

```
Output keys: A, B, C
Scheduler: DaskScheduler
Providers:
....
```
For the dask scheduler, it would be very helpful if we can include which underlying dask scheduler (`get` function) is used, see here: https://scipp.github.io/sciline/generated/classes/sciline.scheduler.DaskScheduler.html.

jokasimr: It was a bit of an experiment ;) I'll change it.

However I don't think the last column should not be "Providers" but rather something like "Input parameters"?
Or do you think it would be better to list all providers in the task graph rather than only the inputs?

Once the user has selected what output they are interested in I think the intermediate providers are kind of uninteresting, while the inputs that contribute to a particular output might be interesting for debugging purposes / understanding what is needed to compute a particular quantitiy.

SimonHeybrock: > Once the user has selected what output they are interested in I think the intermediate providers are kind of uninteresting

Maybe? We can only list the inputs for now.

jokasimr:
New style 
Screenshot from 2024-01-12 10-47-09
Screenshot from 2024-01-12 10-47-24

src/sciline/utils.py
```diff
@@ -16,3 +18,13 @@ def qualname(obj: Any) -> str:
...
+
+
+def keyname(key: Key) -> str:
+    if isinstance(key, Item):
+        return f'{keyname(key.tp)}({keyname(key.label[0].tp)})'
+    args = get_args(key)
+    if len(args):
+        parameters = ', '.join(map(keyname, args))
+        return f'{qualname(key)}[{parameters}]'
+    return qualname(key)
```
SimonHeybrock, line 30: How does this compare to what is used in `visualize`? Can we use the same code?

jokasimr:
New version displaying some more details about the scheduler
Screenshot from 2024-01-12 11-06-23
Screenshot from 2024-01-12 11-06-43

src/sciline/task_graph.py
```diff
@@ -79,3 +110,39 @@ def visualize(
...
+                for key in (
+                    self._keys if isinstance(self._keys, tuple) else [self._keys]
+                )
+            ]
+        )
+        roots = sorted(
+            {
+                escape(keyname(key))
+                for key, (_, requires) in self._graph.items()
+                if len(requires) == 0
```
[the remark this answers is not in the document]
jokasimr: Hm I'm not sure I understand. Is there a conceptual difference between a parameter and a provider without inputs?

src/sciline/utils.py
jokasimr: Yes I think we can.

src/sciline/task_graph.py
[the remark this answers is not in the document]
SimonHeybrock: Yes, a parameter is typically something a user sets, i.e., it would typically be different every time the pipeline is used (and therefore seeing its value is valuable). A provider is more static and would typically not change (and we cannot get its value without calling it, which may be expensive).

SimonHeybrock (approved):

PR #101, opened 2024-01-16
docs: add examples of replacing providers

### LINKED ISSUE #97: Add example of how to continue from computed intermediate result
After computing an intermediate result, we can use `Pipeline.__setitem__` to replace the provider with the intermediate result (as a param):

```python
data = pipeline.compute(Intermediate)
pipeline[Intermediate] = data
result = pipeline.compute(Final)
```

This may not be obvious to users, we should document this in an example.

jokasimr:
Fixes #97 

[replacing-providers.pdf](https://github.com/scipp/sciline/files/13948939/replacing-providers.pdf)


docs/user-guide/index.md
```diff
@@ -8,4 +8,5 @@ maxdepth: 2
 getting-started
 parameter-tables
 generic-providers
+replacing-providers
```
SimonHeybrock, line 11: I suggest we add this in `Recipes`, instead of `User Guide`

docs/user-guide/replacing-providers.ipynb
```diff
@@ -0,0 +1,277 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "86de99b0-3170-45d6-84eb-adbd622af936",
+   "metadata": {},
+   "source": [
+    "# Replacing providers\n",
```
SimonHeybrock, line 8: This is not exactly what the issue was about. I think `insert` and `__setitem__` can be found in the API docs. The issue was about documenting how to *continue from intermediate results*. Someone who wants to do that might not look for "replacing providers" (if they would, then they would already know how to do that). I think we should make this clear in the title, and maybe stick to the first example on this page (linking the `insert` can be useful).

If you want to keep the "replacing provider" example, I'd suggest to make this into another recipe?

jokasimr: Sounds good!

docs/recipes/recipes.ipynb
```diff
@@ -103,6 +103,276 @@
     "Using `bind_and_call` guarantees that the file gets written and that it gets written after the pipeline.\n",
     "The latter prevents providers from accidentally relying on the file."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "60f9b2a8-0557-43da-a8d2-41df2794bd49",
+   "metadata": {},
+   "source": [
+    "## Continue from intermediate results\n",
```
SimonHeybrock, line 112: Hmm, maybe we should split this page into individual notebooks?

docs/recipes/recipes.ipynb
```diff
@@ -103,6 +103,276 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "aa1df625-9f6a-44d3-9169-1c79ff8e647f",
+   "metadata": {
+    "jp-MarkdownHeadingCollapsed": true
+   },
+   "source": [
+    "### Setup\n",
```
SimonHeybrock, line 125: Before the full and working example, should we give a TL;DR?

```python
data = pipeline.compute(CleanData)
pipeline[CleanData] = data
result = pipeline.compute(Result)
```

docs/recipes/recipes.ipynb
```diff
@@ -103,6 +103,276 @@
...
+    "pipeline.get(Result)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "99fe0c69-597b-4384-96de-92b25b6e95cd",
+   "metadata": {},
+   "source": [
+    "But if the cleaned data has already been produced it is unnecessary to \"re-clean\" it, in that case we can proceed directly from the clean data to the compute sum step.\n",
+    "To do this we replace the `CleanData` provider with the data that was loaded and cleaned:"
```
SimonHeybrock, line 210: This is a bit confusing, because the *computation* of `CleanData` comes after the above line, where you say we replace the provider. A novice user might thing the call to `compute` is part of the replacement. How about something like:

Given a pipeline, we may want to compute an intermediate result for inspection:
```python
data = pipeline.compute(CleanData)
```

If later on we wish to compute a result further down the pipeline (derived from `CleanData`), this would cause potentially very costly re-computation of `CleanData`, since Sciline does not perform any caching:
```python
result = pipeline.compute(Result)  # re-computes CleanData
```

To avoid this, we can use `Pipeline.__setitem__` to replace the provider of `CleanData` by the previously computed data:
```python
pipeline[CleanData] = data
result = pipeline.compute(Result)  # uses data as CleanData
```

docs/recipes/recipes.ipynb
```diff
@@ -103,6 +103,276 @@
...
+    "This example shows how to replace a provider in the pipeline using the `Pipeline.insert` method."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f05644be-ac0e-46c2-81a0-a4afe057e411",
+   "metadata": {},
+   "source": [
+    "### Setup\n",
+    "Same setup as in [Continue from intermediate results](#continue-from-intermediate-results)."
```
SimonHeybrock, line 269: Let us avoid this, recipes should be stand-alone.

docs/recipes/recipes.ipynb
jokasimr: Yeah that makes sense

docs/recipes/side-effects-and-file-writing.ipynb
```diff
@@ -1,13 +1,5 @@
 {
  "cells": [
-  {
-   "cell_type": "markdown",
-   "id": "d7608bb4-d7ce-4a36-8021-f1a52ad27ec5",
-   "metadata": {},
-   "source": [
-    "# Recipes"
```
SimonHeybrock, line 8: Need to change section heading levels below, and also all the other files.

docs/recipes/continue-from-intermediate-results.ipynb
```diff
@@ -0,0 +1,164 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "23fe22b8-b59e-4d63-9255-f510bbd8bec7",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pipeline[CleanData] = data\n",
+    "pipeline.compute(Result)"
```
SimonHeybrock, line 139:
```suggestion
    "result = pipeline.compute(Result)"
```
for symmetry with the above?

SimonHeybrock (approved):

SimonHeybrock (approved):

PR #110, opened 2024-01-24
Add domain_types sphinx extension

### LINKED ISSUE #61: Documentation with links to type aliases?
The following is not really a problem of Sciline itself, it does however severely impact usability of documentation of downstream projects.

Sphinx, in particular with https://github.com/tox-dev/sphinx-autodoc-typehints, "expands" type hints to aliases defined with `typing.NewType`. That is, instead of a link to the docs of the alias, we just get a long to `NewType` (which is useless) and the name of the alias.

According to https://github.com/tox-dev/sphinx-autodoc-typehints/issues/216#issuecomment-1026588210 this might be intentional. Can this be changed in the extension as suggested in the comment? Do we need our own extension?

jl-wynen:
Fixes #61

This turns the formatter from https://github.com/scipp/copier_template/pull/124 into a sphinx extension. I am going to add code to use it to the copier template.

### DIFF
--- a/src/sciline/sphinxext/domain_types.py
+++ b/src/sciline/sphinxext/domain_types.py
@@ -5,7 +5,7 @@
 `sphinx-autodoc-typehints <https://github.com/tox-dev/sphinx-autodoc-typehints>`_
 implements formatting for typehints but often results in hard-to-read annotations
 for :class:`typing.NewType`.
-As those are ubiquitous when using Sciline, this extension improves formatting.
+As those may be ubiquitous when using Sciline, this extension improves formatting.
 
 Usage
 -----
--- a/src/sciline/sphinxext/domain_types.py
+++ b/src/sciline/sphinxext/domain_types.py
@@ -33,14 +33,17 @@ Options
   For example, if a type is defined as ``mypackage.types.SomeType``, setting
   ``sciline_domain_types_prefix = 'mypackage'`` results in this type being rendered
   as ``types.SomeType``.
-- ``sciline_domain_types_extra_aliases`` (default ``{}``):
-  ``sciline.sphinxext.domain_types`` can render aliases instead of the regular
-  type names. This is useful, e.g., for abbreviating long types or types defined
-  in internal modules. ``sciline.sphinxext.domain_types`` defines some default
-  aliases for Scipp, e.g., ``{'scipp._scipp.core.DataArray': 'scipp.DataArray', ...}``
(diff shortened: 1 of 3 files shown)


src/sciline/sphinxext/domain_types.py
```diff
@@ -0,0 +1,153 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2024 Scipp contributors (https://github.com/scipp)
+"""A Sphinx extension for improved typehints rendering.
+
+`sphinx-autodoc-typehints <https://github.com/tox-dev/sphinx-autodoc-typehints>`_
+implements formatting for typehints but often results in hard-to-read annotations
+for :class:`typing.NewType`.
+As those are ubiquitous when using Sciline, this extension improves formatting.
```
SimonHeybrock, line 8:
```suggestion
As those may be ubiquitous when using Sciline, this extension improves formatting.
```
`NewType` might just be how we currently use it.

src/sciline/sphinxext/domain_types.py
```diff
@@ -0,0 +1,153 @@
...
+  For example, if a type is defined as ``mypackage.types.SomeType``, setting
+  ``sciline_domain_types_prefix = 'mypackage'`` results in this type being rendered
+  as ``types.SomeType``.
+- ``sciline_domain_types_extra_aliases`` (default ``{}``):
+  ``sciline.sphinxext.domain_types`` can render aliases instead of the regular
+  type names. This is useful, e.g., for abbreviating long types or types defined
+  in internal modules. ``sciline.sphinxext.domain_types`` defines some default
+  aliases for Scipp, e.g., ``{'scipp._scipp.core.DataArray': 'scipp.DataArray', ...}``
+  which means that occurrences of ``scipp._scipp.core.DataArray`` are rendered as
+  ``scipp.DataArray``. Aliases defined in ``sciline_domain_types_extra_aliases``
```
SimonHeybrock, line 42: Sciline is independent of Scipp. Should we keep this out of the project?

SimonHeybrock (commented):
Nice :+1:

src/sciline/sphinxext/domain_types.py
jl-wynen: Then we have to put it into the template to duplicate it across repos. Shouldn't be a problem, but this here seemed more convenient for our target audience.

SimonHeybrock: Is there a problem with putting it into the template? If not, I would suggest to move this there.

Issue #107, opened 2024-01-22
Support for graph operations


jl-wynen:
At the moment, `Pipeline` is largely a black box in terms of the graph it represents. It can build `TaskGraph` objects but only when the user knows suitable nodes to pass to `build` or `get` (as in `tg = p.get(p._providers.keys())` below). We have several issues (open and closed) now that relate to graph operations and could be solved easily if we had such operations:

- #43 can be solved with a graph partitioning algorithm (as suggested in https://github.com/scipp/sciline/issues/43#issuecomment-1876865244)
- #92 Potentially simple to implement with graph traversal functionality.
- Graph visualisation including #73 and existing graphviz converter.
- #80 and #83 are a bit trickier; we would have to build a partial graph and scan it for nodes that are missing an edge (i.e. missing provider for an arg)

NetworkX seems like a good choice for a graph implementation because it provides all basic algorithms out of the box. It becomes trivial to solve #43 as suggested by @YooSunYoung, #92 for several formats, and maybe visualisation (might get removed from NetworkX).

## Current graph traversal

Unfortunately, it is currently cumbersome, expensive, and not technically allowed to traverse graphs. E.g., to build a NetworkX graph, we need something like (note the protected attributes and functions)
```python
import networkx as nx
from sciline.visualize import _format_type as format_type
from sciline import Pipeline

def to_networkx(p: Pipeline) -> nx.DiGraph:
    tg = p.get(p._providers.keys())
    g = nx.DiGraph()
    for res, (provider, args) in tg._graph.items():
        g.add_node(hash(res), label=format_type(res).name)
        for arg in args:
            g.add_node(hash(arg), label=format_type(arg).name)
            g.add_edge(hash(arg), hash(res), label=provider.__qualname__)
    return g
```

## Potential solutions

I talk about NetworkX here, but we can also use a different library or custom graph implementation. The general ideas remain the same.

### 1. Implement `to_networkx` in `Pipeline`

This is the least invasive and easiest to implement. But it would mean that operations that need to inspect the graph need to make a temporary `nx.DiGraph` which is wasteful.

### 2. Represent the internal state of `Pipeline` as a `DiGraph`

NetworkX allows storing arbitrary data in nodes and edges. So we can stuff everything we need into the graph. This would make graph operations first class citizens and would open up more possibilities. (E.g., we can use node metadata to indicate whether a provider is a param or not. Something that is currently difficult (impossible?) to determine.)
[...]

PR #111, opened 2024-01-25
Experiment with building graph from pipeline

SimonHeybrock:
Experiment for #107.

There are conceptual questions regarding how generics should be dealt with. We have not found a consistent and useful formulation so far.


SimonHeybrock:
Ultimately this led to the work in branch `v2-prototype`.

Issue #115, opened 2024-02-02
Earlier detection of missing type hints


jl-wynen:
Currently, this succeeds without error:
```python
def f(x) -> int:
    return x

pl = sl.Pipeline([f])
tg = pl.get(int)
```
But calling `tg.compute(int)` raises `TypeError: f() missing 1 required positional argument: 'x'`. It would be easier to understand what went wrong if the pipeline would raise when the provider is inserted.

Is see no valid use case for a provider like this. In the given form, it will always lead to an exception during `compute`. If there is a default value, it works. But the default can never be overridden by the pipeline. So IMHO, it shouldn't be there at all in a provider.
Note that this is actually used here: https://github.com/scipp/sciline/blob/262fac770ac49de60045a40a019b458271fb269e/src/sciline/pipeline.py#L425 But since `label` can never be provided by the pipeline, the argument can simply be removed from the lambda: `lambda: label` (This does not work completely here because the lambda is defined in a loop and would incorrectly bind the first definition of `label` instead of the one of the current loop iteration.)

SimonHeybrock:
Does #116 fix this?

jl-wynen:
It pushes the detection forward to `pipeline.get`. But not to the constructor.

PR #116, opened 2024-02-02
Support keyword-only arguments

### LINKED ISSUE #76: Support keyword arguments in providers
All arguments are passed positionally. So we cannot use keyword-only arguments in providers.

But since providers are typically not just a thin wrapper but encode some logic themselves, we may want to use them like regular functions. So following our style of requiring keyword arguments in many places, we should allow the same here.

jl-wynen:
Fixes #76

### DIFF
--- a/src/sciline/_provider.py
+++ b/src/sciline/_provider.py
@@ -27,7 +27,7 @@ ToProvider = Callable[..., Any]
 """Callable that can be converted to a provider."""
 
 ProviderKind = Literal[
-    'function', 'parameter', 'series', 'table_label', 'sentinel', 'unsatisfied'
+    'function', 'parameter', 'series', 'table_cell', 'sentinel', 'unsatisfied'
 ]
 """Identifies the kind of a provider, most are used internally."""
 
@@ -80,14 +80,14 @@ class Provider:
         )
 
     @classmethod
-    def table_label(cls, param: Any) -> Provider:
+    def table_cell(cls, param: Any) -> Provider:
         """Construct a provider that returns the label for a table row."""
         return cls(
             func=lambda: param,
             arg_spec=ArgSpec.null(),
-            kind='table_label',
+            kind='table_cell',
             location=ProviderLocation(
-                name=f'table_label({type(param).__name__})', module=_module_name(param)
+                name=f'table_cell({type(param).__name__})', module=_module_name(param)
             ),
         )
 
--- a/src/sciline/_provider.py
+++ b/src/sciline/_provider.py
@@ -3,25 +3,142 @@
 from __future__ import annotations
 
 import inspect
+from dataclasses import dataclass
 from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
     Generator,
+    Literal,
+    Optional,
     TypeVar,
+    Union,
     get_args,
     get_origin,
     get_type_hints,
 )
 
 if TYPE_CHECKING:
-    from .typing import Key, Provider
+    from .typing import Key
+
+
+ToProvider = Callable[..., Any]
+ProviderKind = Literal['function', 'parameter', 'series', 'table', 'sentinel']
 
 
 class UnboundTypeVar(Exception):
     """
     Raised when a parameter of a generic provider is not bound to a concrete type.
     """
+
+
+class Provider:
+    def __init__(
+        self,
+        *,
+        func: ToProvider,
+        arg_spec: ArgSpec,
+        kind: ProviderKind,
+        location: Optional[ProviderLocation] = None,
+    ) -> None:
+        self._func = func
+        self._arg_spec = arg_spec
+        self._kind = kind
+        self._location = (
+            location if location is not None else ProviderLocation.from_function(func)
+        )
+
+    @classmethod
+    def from_function(cls, func: ToProvider) -> Provider:
+        return cls(func=func, arg_spec=ArgSpec.from_function(func), kind='function')
+
+    @classmethod
+    def parameter(cls, param: Any) -> Provider:
+        return cls(
+            func=lambda: param,
+            arg_spec=ArgSpec.null(),
+            kind='parameter',
+            location=ProviderLocation(
+                name=f'param({type(param).__name__})', module='sciline'
+            ),
+        )
+
+    @classmethod
+    def table(cls, param: Any) -> Provider:
+        return cls(
+            func=lambda: param,
+            arg_spec=ArgSpec.null(),
+            kind='table',
+            location=ProviderLocation(
+                name=f'table({type(param).__name__})', module='sciline'
+            ),
+        )
+
+    @classmethod
+    def provide_none(cls) -> Provider:
+        """Provider that takes no arguments and returns None."""
+        return cls(
+            func=lambda: None,
+            arg_spec=ArgSpec.null(),
+            kind='function',
+            location=ProviderLocation(name='provide_none', module='sciline'),
+        )
+
+    @property
+    def func(self) -> ToProvider:
+        return self._func
+
+    @property
+    def arg_spec(self) -> ArgSpec:
+        return self._arg_spec
+
+    @property
+    def kind(self) -> ProviderKind:
+        return self._kind
+
+    @property
+    def location(self) -> ProviderLocation:
+        return self._location
+
+    @property
+    def qualname(self) -> str:
+        # TODO merge with location
+        return getattr(self._func, '__qualname__', self._func.__class__.__qualname__)
+
+    def deduce_key(self) -> Any:
+        if (key := get_type_hints(self._func).get('return')) is None:
+            raise ValueError(
+                f'Provider {self} lacks type-hint for return value or returns NOne.'
+            )
+        return key
+
+    def bind(self, bound: dict[TypeVar, Key]) -> Provider:
+        return Provider(
+            func=self._func, arg_spec=self._arg_spec.bind(bound), kind=self._kind
+        )
+
+    def map_arg_keys(self, transform: Callable[[Key], Key]) -> Provider:
+        return Provider(
+            func=self._func,
+            arg_spec=self._arg_spec.map_keys(transform),
+            kind=self._kind,
+        )
+
+    def __str__(self) -> str:
+        return f"Provider('{self.location.name}')"
+
+    def __repr__(self) -> str:
+        return (
+            f"Provider('{self.location.module}.{self.location.name}', "
+            f"func={self._func})"
+        )
+
+    def call(self, values: dict[Key, Any]) -> Any:
+        """Call the provider with arguments extracted from ``values``."""
+        return self._func(
+            *(values[arg] for arg in self._arg_spec.args),
+            **{key: values[arg] for key, arg in self._arg_spec.kwargs},
+        )
 
 
 class ArgSpec:
@@ -33,18 +150,12 @@ class ArgSpec:
         self._kwargs = kwargs
 
     @classmethod
-    def from_provider(cls, provider: Provider, bound: dict[TypeVar, Key]) -> ArgSpec:
+    def from_function(cls, provider: ToProvider) -> ArgSpec:
         """Parse the argument spec of a provider."""
         hints = get_type_hints(provider)
         signature = inspect.getfullargspec(provider)
-        args = {
-            name: _bind_free_typevars(hints[name], bound=bound)
-            for name in signature.args
-        }
-        kwargs = {
-            name: _bind_free_typevars(hints[name], bound=bound)
-            for name in signature.kwonlyargs
-        }
+        args = {name: hints[name] for name in signature.args}
+        kwargs = {name: hints[name] for name in signature.kwonlyargs}
         return cls(args=args, kwargs=kwargs)
 
     @classmethod
@@ -70,6 +181,10 @@ class ArgSpec:
         yield from self._args.values()
         yield from self._kwargs.values()
 
+    def bind(self, bound: dict[TypeVar, Key]) -> ArgSpec:
+        """Bind concrete types to TypeVars."""
+        return self.map_keys(lambda arg: _bind_free_typevars(arg, bound=bound))
+
     def map_keys(self, transform: Callable[[Key], Key]) -> ArgSpec:
         """Return a new ArgSpec with the keys mapped by ``callback``."""
         return ArgSpec(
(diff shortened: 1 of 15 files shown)


src/sciline/pipeline.py
```diff
@@ -422,12 +406,12 @@ def set_param_table(self, params: ParamTable) -> None:
             for index, label in zip(params.index, values):
                 self._set_provider(
                     Item((Label(tp=params.row_dim, index=index),), param_name),
-                    lambda label=label: label,
+                    _provider_for_constant(label),
```
jl-wynen, line 409: Note the logic in `ArgSpec.from_provider`, implements a hard requirement for not allowing arguments without type hints as per #115. Thus, the old lambda does not work any more.

tests/conftest.py
jl-wynen, line 1: I think it makes sense to expand the use of these fixtures or control the default scheduler globally. As it stands, tests either use dask or the naive scheduler depending on whether dask is installed. This makes it hard to understand how the tests are executed.

SimonHeybrock:
There are a bunch of places in `visualize` and iirc also in the html formatting that check for the internal lambda to detect certain things. That needs updating.

src/sciline/scheduler.py
```diff
@@ -77,7 +77,17 @@ def __init__(self, scheduler: Optional[Callable[..., Any]] = None) -> None:
...
     def get(self, graph: Graph, keys: List[Key]) -> Any:
-        dsk = {tp: (provider, *args) for tp, (provider, args) in graph.items()}
+        from dask.utils import apply
+
+        dsk = {
+            tp: (
+                apply,
+                provider,
+                list(args.args),
+                (dict, [[key, val] for key, val in args.kwargs]),
```
SimonHeybrock, line 87: A comment would be useful, it is not obvious for a reader what is going on here.

tests/pipeline_test.py
```diff
@@ -1208,3 +1208,43 @@ def process(
 def test_html_repr() -> None:
     pipeline = sl.Pipeline([make_int], params={float: 5.0})
     assert isinstance(pipeline._repr_html_(), str)
+
+
+def test_pipeline_keyword_argument_and_param(scheduler: sl.scheduler.Scheduler) -> None:
```
SimonHeybrock, line 1213: Can you name this (and below) with something like `keyword_only` in the test name? I think keyword args have always been supported?

src/sciline/_provider.py
```diff
@@ -0,0 +1,99 @@
...
+            args={name: transform(arg) for name, arg in self._args.items()},
+            kwargs={name: transform(arg) for name, arg in self._kwargs.items()},
+        )
+
+    def call(self, provider: Provider, values: dict[Key, Any]) -> Any:
+        """Call a compatible provider with arguments extracted from ``values``."""
+        return provider(
+            *(values[arg] for arg in self._args.values()),
+            **{key: values[arg] for key, arg in self._kwargs.items()},
+        )
```
SimonHeybrock, line 85: Does this belong here? It seems it is used only by `NaiveScheduler`?

SimonHeybrock (commented):
Looks good! Minor questions (aside from the required fixes for `visualize`, etc.).

src/sciline/_provider.py
jl-wynen: True, I don't like this one either

jl-wynen:
> There are a bunch of places in visualize and iirc also in the html formatting that check for the internal lambda to detect certain things. That needs updating.

I considered also introducing a dedicated `Provider` class. This would encapsulate these checks. E.g., when the class is instantiated, we could specify whether it is a regular provider, parameter, or series provider. Then we don't have to mess with detecting lambdas. Would this be a suitable change here?

SimonHeybrock:
> > There are a bunch of places in visualize and iirc also in the html formatting that check for the internal lambda to detect certain things. That needs updating.
> 
> I considered also introducing a dedicated `Provider` class. This would encapsulate these checks. E.g., when the class is instantiated, we could specify whether it is a regular provider, parameter, or series provider. Then we don't have to mess with detecting lambdas. Would this be a suitable change here?

Sounds reasonable!

src/sciline/task_graph.py
```diff
@@ -7,7 +7,8 @@
 
 from .scheduler import DaskScheduler, NaiveScheduler, Scheduler
 from .typing import Graph, Item
-from .utils import keyname, kind_of_provider
+
+# from .utils import keyname, kind_of_provider
```
SimonHeybrock, line 11: To remove?

src/sciline/utils.py
```diff
@@ -32,11 +32,11 @@ def keyname(key: Union[Key, TypeVar]) -> str:
...
 
-def kind_of_provider(p: Callable[..., Any]) -> ProviderKind:
-    from .pipeline import Pipeline
-
-    if qualname(p) == f'{qualname(Pipeline.__setitem__)}.<locals>.<lambda>':
-        return 'parameter'
-    if qualname(p) == f'{qualname(Pipeline.set_param_table)}.<locals>.<lambda>':
-        return 'table'
-    return 'function'
+# def kind_of_provider(p: Callable[..., Any]) -> ProviderKind:
```
SimonHeybrock, line 35: Remove?

src/sciline/visualize.py
```diff
@@ -85,47 +93,45 @@ def get_subgraph_name(name: str) -> str:
...
+                formatted_p.ret.name,
+                shape='box3d' if formatted_p.ret.collapsed else 'rectangle',
             )
         # Do not draw dummy providers created by Pipeline when setting instances
-        if p_name in (
-            f'{_qualname(Pipeline.__setitem__)}.<locals>.<lambda>',
-            f'{_qualname(Pipeline.set_param_table)}.<locals>.<lambda>',
-            unsatisfied,
-        ):
+        if formatted_p.kind in ('parameter', 'table', 'sentinel'):
```
SimonHeybrock, line 123: What is `sentinel` here? And why are we still using `unsatisfied` above but not here?

tests/pipeline_with_param_table_test.py
```diff
@@ -350,6 +350,7 @@ def test_can_groupby_by_requesting_series_of_series() -> None:
         Param1,
         {1: sl.Series(Row, {0: 4, 1: 5}), 3: sl.Series(Row, {2: 6})},
     )
+    print(list(pl._providers.values()))
```
SimonHeybrock, line 353: remove?

src/sciline/_provider.py
```diff
@@ -3,19 +3,27 @@
...
     get_type_hints,
 )
 
 if TYPE_CHECKING:
-    from .typing import Key, Provider
+    from .typing import Key
+
+
+ToProvider = Callable[..., Any]
+ProviderKind = Literal['function', 'parameter', 'series', 'table', 'sentinel']
```
SimonHeybrock, line 26: Any reason why we use literals instead of, e.g., an enum?

src/sciline/_provider.py
```diff
@@ -24,6 +32,115 @@ class UnboundTypeVar(Exception):
...
+        return cls(func=func, arg_spec=ArgSpec.from_function(func), kind='function')
+
+    @classmethod
+    def parameter(cls, param: Any) -> Provider:
+        return cls(
+            func=lambda: param,
+            arg_spec=ArgSpec.null(),
+            kind='parameter',
+            location=ProviderLocation(
+                name=f'param({type(param).__name__})', module='sciline'
```
SimonHeybrock, line 62: Why is the module `sciline`? Shouldn't we use the module of the type?

src/sciline/_provider.py
```diff
@@ -24,6 +32,115 @@ class UnboundTypeVar(Exception):
...
+            ),
+        )
+
+    @classmethod
+    def table(cls, param: Any) -> Provider:
```
SimonHeybrock, line 67: I guess it is more like a table-cell than a table?

src/sciline/_provider.py
```diff
@@ -24,6 +32,115 @@ class UnboundTypeVar(Exception):
...
+        return cls(
+            func=lambda: param,
+            arg_spec=ArgSpec.null(),
+            kind='table',
+            location=ProviderLocation(
+                name=f'table({type(param).__name__})', module='sciline'
```
SimonHeybrock, line 73: Same question about the module as above.

src/sciline/_provider.py
```diff
@@ -24,6 +32,115 @@ class UnboundTypeVar(Exception):
...
+    def kind(self) -> ProviderKind:
+        return self._kind
+
+    @property
+    def location(self) -> ProviderLocation:
+        return self._location
+
+    @property
+    def qualname(self) -> str:
+        # TODO merge with location
```
SimonHeybrock, line 105: todo

src/sciline/pipeline.py
```diff
@@ -316,19 +327,20 @@ def __init__(
...
+            Either a callable that provides its return value. Its arguments
+            and return value must be annotated with type hints.
+            Or a ``Provider`` object that has been constructed from such a callable.
         """
-        if (key := get_type_hints(provider).get('return')) is None:
-            raise ValueError(f'Provider {provider} lacks type-hint for return value')
-        self._set_provider(key, provider)
+        if not isinstance(provider, Provider):
+            provider = Provider.from_function(provider)
+        self._set_provider(provider.deduce_key(), provider)
```
SimonHeybrock, line 343: Why allow creation of the provider if the key cannot be deduced? Should the check be moved into `from_function`?

src/sciline/pipeline.py
```diff
@@ -406,12 +418,12 @@ def set_param_table(self, params: ParamTable) -> None:
             for index, label in zip(params.index, values):
                 self._set_provider(
                     Item((Label(tp=params.row_dim, index=index),), param_name),
-                    _provider_for_constant(label),
+                    Provider.table(label),
                 )
         for index, label in zip(params.index, params.index):
             self._set_provider(
                 Item((Label(tp=params.row_dim, index=index),), params.row_dim),
```
SimonHeybrock, line 425: Not necessary, but could this clumsy key setup be absorbed into the `Provider` class, similar to the `function` case? `Provider.table` would obviously need more arguments.

src/sciline/_provider.py
jl-wynen: It was like this and I didn't want to change too much. I don't mind either way

src/sciline/pipeline.py
jl-wynen: `from_function` can, and probably should do this check. But we need to check again in `deduce_key` because we cannot deduce non-function provider keys.

src/sciline/pipeline.py
jl-wynen: Would you store the key in the provider? This would duplicate the key as we still need it in the graph / pipeline.

jl-wynen:
I think I addressed all you comments.

src/sciline/_provider.py
```diff
@@ -50,27 +64,30 @@ def __init__(
...
             location=ProviderLocation(
-                name=f'param({type(param).__name__})', module='sciline'
+                name=f'param({type(param).__name__})', module=_module_name(param)
             ),
         )
 
     @classmethod
-    def table(cls, param: Any) -> Provider:
+    def table_row(cls, param: Any) -> Provider:
+        """Construct a provider that returns the label for a table row."""
```
SimonHeybrock, line 84: I think it is not a row that is being returned, just a single cell (i.e., specific row and column).

src/sciline/_provider.py
```diff
@@ -23,7 +24,12 @@
 
 
 ToProvider = Callable[..., Any]
-ProviderKind = Literal['function', 'parameter', 'series', 'table', 'sentinel']
+"""Callable that can be converted to a provider."""
+
+ProviderKind = Literal[
+    'function', 'parameter', 'series', 'table', 'sentinel', 'unsatisfied'
```
SimonHeybrock, line 30: Is 'sentinel' still in use?

src/sciline/_provider.py
jl-wynen: It is a single value. E.g. in https://scipp.github.io/sciline/user-guide/parameter-tables.html#Computing-results-for-series-of-parameters it is a single `int`, i.e., the `RunID`, but not the `Filename`.

src/sciline/_provider.py
jl-wynen: Yes, in https://github.com/scipp/sciline/blob/011f9bfb6722c8348fa98f3d2b08ec9844909372/src/sciline/pipeline.py#L587 and https://github.com/scipp/sciline/blob/011f9bfb6722c8348fa98f3d2b08ec9844909372/src/sciline/pipeline.py#L591

src/sciline/_provider.py
SimonHeybrock: Yes, i.e., rename?

src/sciline/_provider.py
SimonHeybrock: Ok, but it will never make it to, e.g., `visualize`, right?

jl-wynen: It shouldn't.

src/sciline/_provider.py
```diff
@@ -27,7 +27,7 @@
 """Callable that can be converted to a provider."""
 
 ProviderKind = Literal[
-    'function', 'parameter', 'series', 'table', 'sentinel', 'unsatisfied'
+    'function', 'parameter', 'series', 'table_label', 'sentinel', 'unsatisfied'
```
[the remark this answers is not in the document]
jl-wynen: Because this is what it's called in the code: https://github.com/scipp/sciline/blob/f24a62f51bdee6ec4983b0f1beb90bd5c24f54c3/src/sciline/pipeline.py#L421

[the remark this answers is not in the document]
SimonHeybrock: Yes, but shouldn't that be renamed?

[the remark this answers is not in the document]
jl-wynen: Don't know. Pipeline seems to use 'label' and 'index' interchangeably for tables. So we could rename it to index?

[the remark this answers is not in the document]
SimonHeybrock: No, this is two different things. If you rename it to index it would mean an entry of the index column:

- A param table has an "index" column, and one or more data columns. Columns are labeled using a type.
- The index is a bit like an index in a `pandas.DataFrame` (and it also has a column label, the type of the index).
- A cell in the table is identified using an index value, i.e., row index (an index, in the alternate meaning of the word index, i.e., an entry of the index column) and a column label.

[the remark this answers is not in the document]
SimonHeybrock: You renamed the method to `table_cell`, but here it is still `table_label`.

SimonHeybrock (approved):

PR #118, opened 2024-02-07
feat: check constraints on typevars

### LINKED ISSUE #68: Handling ``TypeVar`` with constraints of ``Generic`` providers.
As mentioned in the PR #67,
consistent way of handling constraints of a ``TypeVar`` should be discussed.

## Context
Currently, constraints of ``TypeVar`` do not have any affect.
And existing ``Generic`` provider with ``TypeVar`` does not allow to have another provider with the same ``Generic`` type but with a specific type argument.

For instance following example does not work due to multiple providers:
```python
import sciline as sl
from typing import TypeVar, NewType

A = NewType('A', str)
B = NewType('B', str)
C = NewType('C', str)

T = TypeVar('T', A, B)

class H(sl.Scope[T, str], str):
    pass

def p1(x: T) -> H[T]:
    return f"H[{x}]"

def p2(x: C) -> H[C]:
    return "H[C]"

sl.Pipeline(
    [p1, p2], params={A: 'A', B: 'B', C: 'C'}
).compute(H[C])  # Raises ambiguous provider error.
```

Therefore @jokasimr suggested to allow specific type argument in the ``Generic`` type that are not in the constraints of the ``TypeVar`` type argument.

With that fix in PR #67, the above example would work.

## Discussions
Here is the summary of the discussion:

@SimonHeybrock suggested not to consider any ``Constraints`` and always prioritize a specific type argument to a ``TypeVar`` type argument instead.

No effect of constraints of ``TypeVar`` may be confusing to users as @jokasimr mentioned,
so I suggested to restrict the usage of the generic provider to the certain types that are in the constraints.

i.e. This following example should fail.
```python
S1 = NewType("S1", str)
S2 = NewType("S2", str)
S3 = NewType("S3", str)

S12 = TypeVar("T", S1, S2)

class H(sl.Scope[V, str], str):
    ...

def p12(S12) -> H[S12]
   ...

pl = sl.Pipeline(p12, params={S1: 'S1', S2:'S2', S3:'S3'})
pl.compute(H[S3])  # Should raise provider not found error
```

One concern @SimonHeybrock raised was that the providers, `p12` in this case, may not be used in the broader scope.

My opinion is that if the provider is meant to be flexible enough to be used in other scopes, then the ``TypeVar`` argument should have ``bound`` instead of ``constraints``.
In this way, we can both define ``strict`` providers and ``flexible`` providers.

## Suggested Solutions
Here are the summary of suggested solutions with **conflicts** to be resolved.

-  Allow multiple providers of the same ``Generic`` type only if the type argument constraints are not conflicting.
- ``constraints`` of ``TypeVar`` has no effect and always prioritize another ``Generic`` provider with specific type arguments.
- ``Generic`` provider with ``TypeVar`` can only be bound to the certain type sin its own ``constraints``.

jokasimr:
Fixes #68

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -14,6 +14,7 @@ from typing import (
     List,
     Mapping,
     Optional,
+    Set,
     Tuple,
     Type,
     TypeVar,
@@ -39,6 +40,7 @@ from .param_table import ParamTable
 from .scheduler import Scheduler
 from .series import Series
 from .typing import Graph, Item, Key, Label, get_optional, get_union
+from .utils import keyname
 
 T = TypeVar('T')
 KeyType = TypeVar('KeyType', bound=Key)
@@ -51,21 +53,68 @@ class AmbiguousProvider(Exception):
     """Raised when multiple providers are found for a type."""
 
 
-def _is_compatible_type_tuple(
+def _extract_typevars_from_generic_type(t: type) -> Tuple[TypeVar, ...]:
+    """Returns the typevars that were used in the definition of a Generic type."""
+    if not hasattr(t, '__orig_bases__'):
+        return ()
+    return tuple(
+        chain(*(get_args(b) for b in t.__orig_bases__ if get_origin(b) == Generic))
+    )
+
+
+def _find_all_typevars(t: Union[type, TypeVar]) -> Set[TypeVar]:
+    """Returns the set of all TypeVars in a type expression."""
+    if isinstance(t, TypeVar):
+        return {t}
+    return set(chain(*map(_find_all_typevars, get_args(t))))
+
+
+def _find_bounds_to_make_compatible_type(
+    requested: Key,
+    provided: Key | TypeVar,
+) -> Optional[Dict[TypeVar, Key]]:
+    """
+    Check if a type is compatible to a provided type.
+    If the types are compatible, return a mapping from typevars to concrete types
+    that makes the provided type equal to the requested type.
+    """
+    if provided == requested:
+        ret: Dict[TypeVar, Key] = {}
+        return ret
+    if isinstance(provided, TypeVar):
+        # If the type var has no constraints, accept anything
+        if not provided.__constraints__:
+            return {provided: requested}
+        for c in provided.__constraints__:
+            if _find_bounds_to_make_compatible_type(requested, c) is not None:
+                return {provided: requested}
+    if get_origin(provided) is not None:
+        if get_origin(provided) == get_origin(requested):
+            return _find_bounds_to_make_compatible_type_tuple(
+                get_args(requested), get_args(provided)
+            )
+    return None
+
+
+def _find_bounds_to_make_compatible_type_tuple(
     requested: tuple[Key, ...],
     provided: tuple[Key | TypeVar, ...],
-) -> bool:
+) -> Optional[Dict[TypeVar, Key]]:
     """
-    Check if a tuple of requested types is compatible with a tuple of provided types.
-
-    Types in the tuples must either by equal, or the provided type must be a TypeVar.
+    Check if a tuple of requested types is compatible with a tuple of provided types
+    and return a mapping from type vars to concrete types that makes all provided
+    types equal to their corresponding requested type.
+    If any of the types is not compatible, return None.
     """
-    for req, prov in zip(requested, provided):
-        if isinstance(prov, TypeVar):
-            continue
-        if req != prov:
-            return False
-    return True
+    union: Dict[TypeVar, Key] = {}
+    for bound in map(_find_bounds_to_make_compatible_type, requested, provided):
+        # If no mapping from the type-var to a concrete type was found,
+        # or if the mapping is inconsistent,
+        # interrupt the search and report that no compatible types were found.
+        if bound is None or any(k in union and union[k] != bound[k] for k in bound):
+            return None
+        union.update(bound)
+    return union
 
 
 def _find_all_paths(
@@ -494,6 +543,7 @@ class Pipeline:
         self, tp: Union[Type[T], Item[T]], handler: Optional[ErrorHandler] = None
     ) -> Tuple[Provider, Dict[TypeVar, Key]]:
         handler = handler or HandleAsBuildTimeException()
+        explanation: List[str] = []
         if (provider := self._providers.get(tp)) is not None:
             return provider, {}
         elif (origin := get_origin(tp)) is not None and (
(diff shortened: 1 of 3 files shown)


tests/pipeline_test.py
```diff
@@ -1208,3 +1208,23 @@ def process(
...
+    def process(value: T) -> MyType[T]:
+        return MyType(value)
+
+    pipeline = sl.Pipeline((process,))
+    pipeline[str] = 'abc'
+    pipeline[int] = 123
+
+    pipeline.get(MyType[int])
+
+    with pytest.raises(sl.handler.UnsatisfiedRequirement):
```
SimonHeybrock, line 1229: Is this the correct exception type?

jokasimr: I think so, we're requiring a `MyType[str]` but there is no provider that defines that quantity.

What exception type would you expect there?

SimonHeybrock: My point is that `MyType[str]` is invalid, whether a provider exists or not.

src/sciline/pipeline.py
```diff
@@ -503,12 +532,27 @@ def _get_provider(
...
+            #
+            # Contrived example:
+            #  We are requiring Person[Likes[Green], Green] and find the following
+            #  matching providers to choose from:
+            #    Person[Likes[~Color], ~Color] <- matches if we set ~Color = Green
+            #    Person[~Preference, Green] <- matches if we
+            #                                        set ~Preference = Likes[Green]
+            #
+            # Currently the second will be chosen because only one of the arguments
+            # is not concrete while for the first type both arguments are not concrete.
```
SimonHeybrock, line 552: Missing a test for this?

src/sciline/pipeline.py
```diff
@@ -518,12 +562,7 @@ def _get_provider(
             ]
 
             if len(matches) == 1:
-                args, provider = matches[0]
-                bound = {
-                    arg: req
-                    for arg, req in zip(args, requested)
-                    if isinstance(arg, TypeVar)
-                }
+                args, provider, bound = matches[0]
```
SimonHeybrock, line 565: `args` unused?

src/sciline/pipeline.py
```diff
@@ -56,21 +56,50 @@ class AmbiguousProvider(Exception):
...
     provided: tuple[Key | TypeVar, ...],
-) -> bool:
+) -> Optional[Mapping[TypeVar, Key]]:
     """
     Check if a tuple of requested types is compatible with a tuple of provided types.
-
-    Types in the tuples must either by equal, or the provided type must be a TypeVar.
+    If the types are compatible, return a mapping from type vars to concrete types
+    that makes all provided types equal to their corresponding requested type.
+    If any of the types are not compatible, return None.
```
SimonHeybrock, line 95: Sounds like the function should be renamed that?

tests/pipeline_test.py
```diff
@@ -1208,3 +1208,36 @@ def process(
...
+
+    def p2(value: S) -> N[S]:
+        return N(value)
+
+    pipeline = sl.Pipeline((p1, p2))
+    pipeline[str] = 'abc'
+    pipeline[int] = 123
+
+    pipeline.get(M[int])
+    pipeline.get(N[M[int]])
```
SimonHeybrock, line 1237: I think I had written a less contrived example in the related issue, can we add that as an additional simpler test?

src/sciline/pipeline.py
jokasimr: Yes I'll add a test for this. But I think we also need to decide what behavior we want there. I'm leaning towards changing it to count the number of type-vars in the whole type expression (the length of `bound`). In the above example that would lead to an `AmbiguousProviderError` because both expressions have the same number of type-vars. This seems more appropriate because I don't think that it's obvious that one provider is more specialized than the other.

What do you think?

SimonHeybrock: Yes, we should raise.

tests/pipeline_test.py
jokasimr: I copied the test you wrote in the issue, the only difference is that I added the nested case. Would it be better if I split the not-nested and the nested case into two tests?

jokasimr: Well, it does test one thing, but in two different cases, which is common pratice...

But I'll split it into two separate tests if you think that's better in this case.

tests/pipeline_test.py
SimonHeybrock: Sure, and in general we check this before constructing the graph. But I still think we should raise a different exception type?

src/sciline/pipeline.py
```diff
@@ -572,15 +572,26 @@ def _get_provider(
...
+                                    keyname(origin),
+                                    ' have constraints ',
+                                    (
+                                        {
+                                            keyname(tv): tuple(
+                                                map(keyname, tv.__constraints__)
+                                            )
+                                            for tv in typevars_in_expression
+                                        }
+                                    ),
```
jokasimr, line 591: This adds an note to the error message that looks something like this:
```python
sciline.handler.UnsatisfiedRequirement: ('No provider found for type', module.N[module.M[str], float], "Note that the type variables ('~T', '~S') of module.N have constraints {'~T': ('int', 'float'), '~S': ('M[int]', 'M[float]')}")
```
It's not very elegant. 
I think it somewhat obscures the real error message which is that no provider was found for the requested type. But it does kind of give you the information you need to understand what went wrong. So maybe it's still useful.

jokasimr: A bit better, now it says:
```python
sciline.handler.UnsatisfiedRequirement: ('No provider found for type', module.N[module.M[str], float], "Note that N[~S, ~T] has constraints {'~S': ('M[int]', 'M[float]'), '~T': ('int', 'float')}")
```

SimonHeybrock:
Note the merge conflicts

tests/pipeline_test.py
```diff
@@ -1208,3 +1208,124 @@ def process(
...
+        return M(value)
+
+    pipeline = sl.Pipeline((p1,))
+    pipeline[str] = 'abc'
+    pipeline[int] = 123
+
+    pipeline.get(M[int])
+
+    with pytest.raises(sl.handler.UnsatisfiedRequirement):
+        pipeline.get(M[str])  # type: ignore[type-var]
```
SimonHeybrock, line 1230: Can we define `M` differently (using a different `T`) such that `mypy` passes on this?

jokasimr: Yeah we could, but I think it would make the test more confusing. The point of the test is that we're trying to request a type that is outside of the constraints of `T`, I think the `#ignore` helps to communicate that.

SimonHeybrock: On the contrary! Remember how *I* was confused, because `M[str]` is illegal. But you then went on to explain to be that what matters is what the provider can provide. I think you must make `M[str]` legal, but keep the provider as it is, i.e., it cannot provider `M[str]`.

SimonHeybrock (approved):

PR #120, opened 2024-02-07
Detect missing typehint early

### LINKED ISSUE #115: Earlier detection of missing type hints

jl-wynen:
Fixes #115

### DIFF
--- a/src/sciline/_provider.py
+++ b/src/sciline/_provider.py
@@ -125,7 +125,7 @@ class Provider:
         """Attempt to determine the key (return type) of the provider."""
         if (key := get_type_hints(self._func).get('return')) is None:
             raise ValueError(
-                f'Provider {self} lacks type-hint for return value or returns NOne.'
+                f'Provider {self} lacks type-hint for return value or returns None.'
             )
         return key
 
@@ -175,8 +175,13 @@ class ArgSpec:
         """Parse the argument spec of a provider."""
         hints = get_type_hints(provider)
         signature = inspect.getfullargspec(provider)
-        args = {name: hints[name] for name in signature.args}
-        kwargs = {name: hints[name] for name in signature.kwonlyargs}
+        try:
+            args = {name: hints[name] for name in signature.args}
+            kwargs = {name: hints[name] for name in signature.kwonlyargs}
+        except KeyError:
+            raise ValueError(
+                f'Provider {provider} lacks type-hint for arguments.'
+            ) from None
         return cls(args=args, kwargs=kwargs)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


src/sciline/_provider.py
```diff
@@ -175,8 +175,13 @@ def from_function(cls, provider: ToProvider) -> ArgSpec:
...
         hints = get_type_hints(provider)
         signature = inspect.getfullargspec(provider)
-        args = {name: hints[name] for name in signature.args}
-        kwargs = {name: hints[name] for name in signature.kwonlyargs}
+        try:
+            args = {name: hints[name] for name in signature.args}
+            kwargs = {name: hints[name] for name in signature.kwonlyargs}
+        except KeyError:
+            raise ValueError(
+                f'Provider {provider} lacks type-hint for arguments.'
```
SimonHeybrock, line 183:
```suggestion
                f'Provider {provider} lacks type-hint for arguments or return value.'
```

src/sciline/_provider.py
```diff
@@ -175,8 +175,13 @@ def from_function(cls, provider: ToProvider) -> ArgSpec:
         """Parse the argument spec of a provider."""
         hints = get_type_hints(provider)
         signature = inspect.getfullargspec(provider)
-        args = {name: hints[name] for name in signature.args}
-        kwargs = {name: hints[name] for name in signature.kwonlyargs}
+        try:
+            args = {name: hints[name] for name in signature.args}
+            kwargs = {name: hints[name] for name in signature.kwonlyargs}
```
SimonHeybrock, line 180: So why did this not raise before? This looks the same to me (apart from a better exception message)?

src/sciline/_provider.py
jl-wynen: No, that is handled separately

SimonHeybrock (approved):

Issue #92, opened 2024-01-05
save pipeline structure with parameters for reproduction


dschick:
Hi scliline team,

First of all, many thanks for this great package. We were currently thinking of designing a similar system for pipeline data processing based on `xarray` data containers and luckily found your work before writing a line of code.

In some of our data analysis tasks, we have some rather expensive `produces` (e.g. phase retrieval methods for X-ray holography) and in addition to the actual result of a pipeline, we would also like to save *how* we got there.

Obviously, we could save the list of producers and parameters on our own, but how about a dedicated method of the `pipeline` class similar to `visualize()`, which could return not only the structure of the graph but also the actual parameter values?

In addition to the names of the producers, one could also think of saving their source code and/or a hash of it for full reproducibility.
This feature could be extremely helpful during beamtimes, when code is often changed during online analysis.

Best

Daniel

jl-wynen:
This was on our list of requirements when designing Sciline but it had a low priority lately. So thanks for the reminder!

As you said, we will likely have to write the graph and parameter values. There are some open questions, though:
- What graph format do we use? Are dot files (which `visualize` effectively uses) enough? This may link to the discussion in #43.
- Do we write the parameters into a separate file? If so, what format? And keep in mind that this needs to work for all parameter types, scipp, builtin, xarray, etc.

> In addition to the names of the producers, one could also think of saving their source code and/or a hash of it for full reproducibility.

This is an interesting idea. But it would be an incomplete solution because providers typically call additional functions and we couldn't reasonably write their source code or hash, too.

In our case, we expect to have a script or Jupyter notebook that defines the graph and possibly some specialised providers as well as one or more packages that define most providers. My assumption was that we at least write the precise version of al relevant packages (or a full `pip freeze` or `conda list`). And, at least for code controlled by us, the full script or notebook that defines the pipeline. Those files can then be archived in SciCat together with the processed data.
But I admit that this requires some work from the pipeline author and only really works when we can associate all files with each other with a catalogue like SciCat.

dschick:
It's great that you have a similar interest here :)
And obviously, your thoughts are already more advanced than mine.

I'll be happy to discuss this further at any point.
Feel free to close this issue for now.

Best

Daniel

SimonHeybrock:
My idea so far was to store the *graph* in a Sciline-independent manner. "Producers" and "Parameters" are strictly speaking an implementation detail of Sciline, so one would not want to rely on this for long-term archiving of data, FAIR data, ...

[...]

PR #124, opened 2024-02-13
JSON serializer for task graphs

jl-wynen:
First step of #92

Please start with the new notebook to read about the reasoning behind the format and see examples.


docs/user-guide/index.md
```diff
@@ -8,4 +8,5 @@ maxdepth: 2
 getting-started
 parameter-tables
 generic-providers
+serialization
```
SimonHeybrock, line 11: Does not match page title

src/sciline/__init__.py
```diff
@@ -10,7 +10,7 @@
     __version__ = "0.0.0"
 
 from . import scheduler
-from ._provider import UnboundTypeVar
+from ._provider import ArgSpec, Provider, ProviderKind, ProviderLocation, UnboundTypeVar
```
SimonHeybrock, line 13: Why are these top-level now?

SimonHeybrock (commented):
Before a complete review, some general questions/concerns:

- Why are data nodes serialized? I thought we just want to serialize the task graph (which may eventually include serializing the *input* data, or a hash thereof).
- It is not clear why we would want special handling for param tables. Is the only reason a potential desire to reconstruct the same `TaskGroup` from the JSON? Otherwise I would feel that the special handling just overcomplicates things. I do not really see a need for handling this any different from "regular" providers.

jl-wynen:
> Why are data nodes serialized? I thought we just want to serialize the task graph (which may eventually include serializing the input data, or a hash thereof).

Why does `visualize` show them? Without the data nodes, it becomes harder to understand what a provider returns and how that relates to arguments of other providers. We would basically have to write the signature of providers into the nodes.

> It is not clear why we would want special handling for param tables. Is the only reason a potential desire to reconstruct the same TaskGroup from the JSON? Otherwise I would feel that the special handling just overcomplicates things. I do not really see a need for handling this any different from "regular" providers.

What is a 'TaskGroup'?

table providers are handled the same as others. It is the data nodes that have special handling because they encode which table entry they represent. If this wasn't the case, we would have a lot of duplicate, e.g., `float(int)` nodes in the graph.

src/sciline/__init__.py
jl-wynen: Leftover from a helper function that I removed again.

SimonHeybrock:
> > Why are data nodes serialized? I thought we just want to serialize the task graph (which may eventually include serializing the input data, or a hash thereof).
> 
> Why does `visualize` show them? Without the data nodes, it becomes harder to understand what a provider returns and how that relates to arguments of other providers. We would basically have to write the signature of providers into the nodes.

`visualize` is not about provenance. I am not saying serializing the result types is a bad idea, but then I thought it should be part of the `'compute'` nodes, not represented as `'data'` nodes. But if we would distinguish between `'input'` nodes (which may get serialized values eventually) and `'data'` nodes (values never serialized) that would be equivalent.


> > It is not clear why we would want special handling for param tables. Is the only reason a potential desire to reconstruct the same TaskGroup from the JSON? Otherwise I would feel that the special handling just overcomplicates things. I do not really see a need for handling this any different from "regular" providers.
> 
> What is a 'TaskGroup'?

`TaskGraph`

> table providers are handled the same as others. It is the data nodes that have special handling because they encode which table entry they represent. If this wasn't the case, we would have a lot of duplicate, e.g., `float(int)` nodes in the graph.

Not an issue if my first item is taken into account.

jl-wynen:
> visualize is not about provenance.

This doesn't answer my question. Why is it useful to show the types as separate nodes in visualize but not for provenance tracking?

> But if we would distinguish between 'input' nodes (which may get serialized values eventually) and 'data' nodes (values never serialized) that would be equivalent.

What would be equivalent?

> > table providers are handled the same as others. It is the data nodes that have special handling because they encode which table entry they represent. If this wasn't the case, we would have a lot of duplicate, e.g., float(int) nodes in the graph.

> Not an issue if my first item is taken into account.

Well, the special handling would be moved from data to provider nodes. But it would still be there.

SimonHeybrock:
> > visualize is not about provenance.
> 
> This doesn't answer my question. Why is it useful to show the types as separate nodes in visualize but not for provenance tracking?

Not sure why you don't see having the types in the visualization is useful. But computation wise, the graph of providers provides a full definition.

> > But if we would distinguish between 'input' nodes (which may get serialized values eventually) and 'data' nodes (values never serialized) that would be equivalent.
> 
> What would be equivalent?

Adding an extra attribute in a compute node saying "I am computing type X" is equivalent to adding a data node between this compute node and all dependent compute nodes.

> > > table providers are handled the same as others. It is the data nodes that have special handling because they encode which table entry they represent. If this wasn't the case, we would have a lot of duplicate, e.g., float(int) nodes in the graph.
> 
> > Not an issue if my first item is taken into account.
> 
> Well, the special handling would be moved from data to provider nodes. But it would still be there.

Why, how? I don't see the need for special handling. You'd just have the same compute node N times (all with a unique `id`).

jl-wynen:
> Why, how? I don't see the need for special handling. You'd just have the same compute node N times (all with a unique id).

And identical return types. Or not easily computer-readable return types (Without special handling, the serialized type becomes, e.g., `"builtins.float(builtins.int:0)"` and that still has special handling in `key_full_qualname`)

SimonHeybrock:
> > Why, how? I don't see the need for special handling. You'd just have the same compute node N times (all with a unique id).
> 
> And identical return types. Or not easily computer-readable return types (Without special handling, the serialized type becomes, e.g., `"builtins.float(builtins.int:0)"` and that still has special handling in `key_full_qualname`)

What is the problem with identical return types? You will get N data nodes (all with a unique id).

jl-wynen:
> > > Why, how? I don't see the need for special handling. You'd just have the same compute node N times (all with a unique id).
> > 
> > 
> > And identical return types. Or not easily computer-readable return types (Without special handling, the serialized type becomes, e.g., `"builtins.float(builtins.int:0)"` and that still has special handling in `key_full_qualname`)
> 
> What is the problem with identical return types? You will get N data nodes (all with a unique id).

Not when you merge data and provider nodes. And even if not, you get a bunch of nodes that are identical up to the id. So you know they represent different things, but cannot tell what they represent.

SimonHeybrock:
> > > > Why, how? I don't see the need for special handling. You'd just have the same compute node N times (all with a unique id).
> > > 
> > > 
> > > And identical return types. Or not easily computer-readable return types (Without special handling, the serialized type becomes, e.g., `"builtins.float(builtins.int:0)"` and that still has special handling in `key_full_qualname`)
> > 
> > 
> > What is the problem with identical return types? You will get N data nodes (all with a unique id).
> 
> Not when you merge data and provider nodes.

I think that discussion is orthogonal and does not change anything.

> And even if not, you get a bunch of nodes that are identical up to the id. So you know they represent different things, but cannot tell what they represent.

So? I thought this is about provenance. If you look at an arbitrary non-param-table graph you also have no idea what it represents, without knowing which input data was used, do you?

jl-wynen:
> So? I thought this is about provenance. If you look at an arbitrary non-param-table graph you also have no idea what it represents, without knowing which input data was used, do you?

But the graph has all graph-related info. Yes, parameters are missing, this was part of the requirements for this task. But with tables, even the graph is incompletely specified:
```mermaid
graph LR
A1("A(Label)") --> B1("B(Label)") --> C
A2("A(Label)") --> B2("B(Label)") --> C
A1("A(Label)") --> D
```
Which part of the param table is used by D and which by C?

To be honest, I don't fully understand your objection. Which part of the information in teh graph should be omitted?

SimonHeybrock:
> But the graph has all graph-related info. Yes, parameters are missing, this was part of the requirements for this task. But with tables, even the graph is incompletely specified:
> [...]
> Which part of the param table is used by D and which by C?

The value "1" from the table is used by D, both "1" and "2" are used by C? "1" and "2" are "input" nodes.

```mermaid
graph LR
1-->A1
2-->A2
A1("A") --> B1("B") --> C
A2("A") --> B2("B") --> C
A1("A") --> D
```

docs/user-guide/provenance.ipynb
```diff
@@ -0,0 +1,279 @@
...
+   "id": "279948ef-67ac-4434-9725-7b9b380c48ed",
+   "metadata": {},
+   "source": [
+    "### Reproducibility\n",
+    "\n",
+    "The JSON format used here was chosen as a simple, future-proof format for task graphs.\n",
+    "It can, however,  only capture part of the actual pipeline.\n",
+    "For example, it only shows the structure of the graph and contains the names of functions and types.\n",
+    "But it does not encode the implementation of those functions or types.\n",
+    "Thus, the graph can only be correctly reconstructed in an environment that contains the same software that was used to write the graph.\n",
```
nvaytet, line 204: Would be nice to show an example of loading the json, building a new task graph from it and computing the result?
But I guess you've only implemented the saving of graphs, not loading them?

src/sciline/visualize.py
```diff
@@ -83,18 +83,16 @@ def to_graphviz(
...
 
 def _to_subgraphs(graph: FormattedGraph) -> Dict[str, FormattedGraph]:
-    def get_subgraph_name(name: str) -> str:
-        sgname = name.split('[')[0]
-        if sgname == 'Series':
+    def get_subgraph_name(name: str, kind: str) -> str:
+        if kind == 'series':
             # Example: Series[RowId, Material[Country]] -> RowId, Material[Country]
             return name.partition('[')[-1].rpartition(']')[0]
-        return sgname
```
nvaytet, line 91: This used to return the first name after splitting with `[`, but is now returning the whole thing?
Is that intentional?

tests/serialize/json_test.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def test_serialize_validate_schema_param_table() -> None:
+    pl = sl.Pipeline([as_float])
+    pl.set_param_table(sl.ParamTable(str, {int: [3, -5]}))
+    graph = pl.build(sl.Series[str, float], handler=sl.HandleAsBuildTimeException())
+    tg = TaskGraph(graph=graph, keys=str)
+    res = tg.serialize()
+    schema = json_schema()
+    jsonschema.validate(res, schema)
```
nvaytet, line 254: Along the lines of my first comment, I think it would be good if we could have something a bit like a round-trip test where we save the graph, then load it again and compute something with it and check that it's the same as what the original graph gives?

But this implies we have something that can load saved graphs...

jl-wynen:
> > But the graph has all graph-related info. Yes, parameters are missing, this was part of the requirements for this task. But with tables, even the graph is incompletely specified:
> > [...]
> > Which part of the param table is used by D and which by C?
> 
> The value "1" from the table is used by D, both "1" and "2" are used by C? "1" and "2" are "input" nodes.
> 
> ```mermaid
> graph LR
> 1-->A1
> 2-->A2
> A1("A") --> B1("B") --> C
> A2("A") --> B2("B") --> C
> A1("A") --> D
> ```

Those input nodes correspond to the `p_table_cell` nodes in the PR and are not regular parameter nodes. So would you preserve those but remove the index information from all dependent nodes? This would mean that you have to walk the graph to find that information.

docs/user-guide/provenance.ipynb
jl-wynen: I considered writing a loader. But that needs more info than is in the graph right now (parameters, package versions). And it's non-trivial because it needs some importlib finagling, so I wouldn't necessarily show it in the docs.

src/sciline/visualize.py
jl-wynen: No, it's an oversight. It means that before all subgraphs that return `Series[*]` were merged. Is this intentional?

SimonHeybrock:
> > > But the graph has all graph-related info. Yes, parameters are missing, this was part of the requirements for this task. But with tables, even the graph is incompletely specified:
> > > [...]
> > > Which part of the param table is used by D and which by C?
> > 
> > 
> > The value "1" from the table is used by D, both "1" and "2" are used by C? "1" and "2" are "input" nodes.
> > 
> >   ```mermaid
> > graph LR
> > 1-->A1
> > 2-->A2
> > A1("A") --> B1("B") --> C
> > A2("A") --> B2("B") --> C
> > A1("A") --> D
> > ```
> >   
> >     
> >   
> >   
> > 
> >   
> > 
> >       graph LR
> > 1-->A1
> > 2-->A2
> > A1("A") --> B1("B") --> C
> > A2("A") --> B2("B") --> C
> > A1("A") --> D
> 
> Those input nodes correspond to the `p_table_cell` nodes in the PR and are not regular parameter nodes. So would you preserve those but remove the index information from all dependent nodes? This would mean that you have to walk the graph to find that information.

Yes, absolutely! Remember that Sciline's parameter tables are just a *convenient way of building task graphs*. They do however have little to do with the resulting task graph and actual computation. As this task is about provenance and reproducibility (independent of Sciline), we have to avoid exposing Sciline "implementation details" in the JSON representation.

In fact I would be tempted to argue that a graph that was made without parameter tables "by hand" should result in the same JSON (modulo something related to the provider that makes the series).

jl-wynen:
Implemented a new schema with these changes:
- no support for param tables
- merged data and provider nodes
- renamed type -> out
- function nodes now encode the order of arguments

src/sciline/serialize/_json.py
```diff
@@ -0,0 +1,146 @@
...
+
+    node = {
+        'id': node_id,
+        'kind': 'function',
+        'label': provider_name(provider),
+        'function': provider_full_qualname(provider),
+        'out': key_full_qualname(key),
+        'args': args,
+        'kwargs': kwargs,
+    }
```
SimonHeybrock, line 110: Without having looked at everything, I think this looks much better now (with the Sciline-specifics removed) 👍 

Question: With the `out` as part of the node, we make all dependent nodes linked to the `id` of the provider. Maybe that is not so great, and we should have an `id` for the data, i.e., re-introduce "data" nodes so they can have their own `id`? Or do you think it does not matter?

jl-wynen: I had a long discussion with Max about this. Separating data and functions is in principle cleaner. It would even remove the need for the parameter kind. And if we ever need to attach attributes to data, we could do so without affecting function nodes.

However, it makes the graph harder to read and the whole process a little more complicated. So the thought was to go with what I implemented now and see how that goes and change later if need be. We still have a little time to figure things out.

SimonHeybrock: > Separating data and functions is in principle cleaner.

I thought it would be cleaner especially since we also have input nodes already, a special kind of data node.

> However, it makes the graph harder to read

By whom?

> and the whole process a little more complicated.

Wouldn't it simply mean that providers need to add two nodes, the compute node and the output data node?

jl-wynen: > I thought it would be cleaner especially since we also have input nodes already, a special kind of data node.

This is what I meant by 

> > It would even remove the need for the parameter kind

> > However, it makes the graph harder to read
>
> By whom?
>
> > and the whole process a little more complicated.
>
> Wouldn't it simply mean that providers need to add two nodes, the compute node and the output data node?

By humans because they need to follow two edges to find how data relates to each other instead of one. And by computers because they need to take into account the two fundamentally different node types. E.g., for reconstructing a task graph, the reader must extract pairs of nodes for each task instead of just one.

For writing, this is not that much more complicated and well contained.

jl-wynen: > But they are connected through a direct edge? Is it actually complicated enough to warrant combining compute and data nodes?

No. This makes a miniscule difference on all points unless we foresee adding more fields to data nodes.
I'm happy to change back. I mainly went this way because you objected to having separate notes initially.

SimonHeybrock: > I mainly went this way because you objected to having separate notes initially.

I did not object, I mainly asked about it! And my main concern there was the discrepancy between inputs (then stored as functions) and data nodes.

It seems Aiida has data nodes? https://aiida.readthedocs.io/projects/aiida-core/en/latest/howto/visualising_graphs.html

jl-wynen: Yes, they do. This links into what I hinted at with storing extra metadata in data nodes. If we, e.g., stored intermediate results in a database or in files, we could link to them here, kind of like Aiida does.

SimonHeybrock (approved):
I think this looks good now.

PR #126, opened 2024-02-14
Add TaskGraph.keys

jl-wynen:
Needed by https://github.com/scipp/essreflectometry/issues/6

I went with `TaskGraph.keys` instead of `TaskGraph.nodes` in order to avoid the discussion of whether data and provider nodes are conceptually separate or the same node, see #124.

### DIFF
--- a/src/sciline/task_graph.py
+++ b/src/sciline/task_graph.py
@@ -70,11 +70,11 @@ class TaskGraph:
         self,
         *,
         graph: Graph,
-        keys: Union[type, Tuple[type, ...], Item[T], Tuple[Item[T], ...]],
+        targets: Union[type, Tuple[type, ...], Item[T], Tuple[Item[T], ...]],
         scheduler: Optional[Scheduler] = None,
     ) -> None:
         self._graph = graph
-        self._keys = keys
+        self._keys = targets
         if scheduler is None:
             try:
                 scheduler = DaskScheduler()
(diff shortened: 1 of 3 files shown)


src/sciline/task_graph.py
```diff
@@ -113,6 +113,19 @@ def compute(
         else:
             return self._scheduler.get(self._graph, [keys])[0]
 
+    def keys(self) -> Generator[Key, None, None]:
+        """Iterate over all keys of the graph.
+
+        Yields all keys, i.e., the types of values that can be computed or are
+        provided as parameters.
```
SimonHeybrock, line 120: Erm, this is odd, because we have `self._keys`, and the `keys` passed in the `__init__`, with a quite different meaning from what you implemented here.

jl-wynen: Renamed the args in other methods to `targets` to match the name in `transform_coords`.

src/sciline/task_graph.py
```diff
@@ -103,15 +103,28 @@ def compute(
         If ``keys`` is a single type, returns the single result that was computed.
         If ``keys`` is a tuple of types, returns a dictionary with type as keys
```
SimonHeybrock, line 104: Docstring would need an update given the rename.

SimonHeybrock (approved):

PR #128, opened 2024-02-15
Support decorated providers (again)

### LINKED ISSUE #127: Support decorated providers
Currently, providers with decorators are broken:
```python
import sciline as sl
import functools

def deco(f):
    @functools.wraps(f)
    def impl(*args, **kwargs):
        return f(*args, **kwargs)
    return impl

@deco
def foo(x: int) -> str:
    return str(x)

pl = sl.Pipeline([foo], params={int: 3})
pl.compute(str)
```
raises 
```
TypeError: foo() missing 1 required positional argument: 'x'
```

I think this was introduced in #116 because
```python
import functools
import inspect
from typing import get_type_hints

def deco(f):
    @functools.wraps(f)
    def impl(*args, **kwargs):
        return f(*args, **kwargs)
    return impl

@deco
def foo(x: int) -> str:
    return str(x)

print(get_type_hints(foo))
# -> {'x': <class 'int'>, 'return': <class 'str'>}
print(inspect.getfullargspec(foo))
# -> FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'return': <class 'str'>})
```
So using `getfullargspec` is not enough, we also need to use `get_type_hints`.

Unfortunately, this breaks keyword-only args:
```python
@deco
def foo(x: int, *k: float) -> str:
    return str(x)

print(get_type_hints(foo))
# -> {'x': <class 'int'>, 'k': <class 'float'>, 'return': <class 'str'>}
print(inspect.getfullargspec(foo))
# -> FullArgSpec(args=[], varargs='args', varkw='kwargs', defaults=None, kwonlyargs=[], kwonlydefaults=None, annotations={'return': <class 'str'>})
```

jl-wynen:
Fixes #127

### DIFF
--- a/docs/recipes/applying-decorators.ipynb
+++ b/docs/recipes/applying-decorators.ipynb
@@ -1,0 +1,135 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "f7e2f743-54a3-4031-bb19-f2bd96d05de6",
+   "metadata": {},
+   "source": [
+    "# Applying decorators\n",
+    "\n",
+    "When using decorators with providers, care must be taken to allow Sciline to recognize the argument and return types.\n",
+    "This is done easiest with [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps).\n",
+    "The following decorator can be safely applied to providers:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "30ef6849-606e-4a32-aa75-e493bfdb0fcd",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import functools\n",
+    "from typing import Any, Callable, TypeVar\n",
+    "\n",
+    "import sciline\n",
+    "\n",
+    "R = TypeVar('R')\n",
+    "\n",
+    "\n",
+    "def deco(f: Callable[..., R]) -> Callable[..., R]:\n",
+    "    @functools.wraps(f)\n",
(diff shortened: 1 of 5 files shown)


YooSunYoung:
I added more tests and opened a PR into this branch: #132

jl-wynen:
#132 does not look like it would block this PR because it doesn't test anything new.

YooSunYoung:
No it won't... just wanted to target to this branch cuz it needed this update.

YooSunYoung (approved):

docs/recipes/applying-decorators.ipynb
```diff
@@ -0,0 +1,135 @@
...
+    "\n",
+    "For Python 3.10+, the decorator itself can be type-annoted like this:\n",
+    "\n",
+    "```python\n",
+    "from typing import ParamSpec, TypeVar\n",
+    "\n",
+    "P = ParamSpec('P')\n",
+    "R = TypeVar('R')\n",
+    "\n",
+    "def deco(f: Callable[P R]) -> Callable[P, R]:\n",
```
SimonHeybrock, line 100: Is there a comma missing here?

SimonHeybrock (commented):
I suppose we should make a bugfix release for this?

docs/recipes/applying-decorators.ipynb
jl-wynen: See #134

jl-wynen:
> I suppose we should make a bugfix release for this?

Yes

PR #132, opened 2024-02-16
More tests on ArgSpec

YooSunYoung:
Maybe we could have have some direct/independent tests on ``ArgSpec``.


tests/_provider_test.py
jl-wynen, line 1: Does this test anything that isn't already covered by other tests?

YooSunYoung: If we only use ``Pipeline`` to test everything, it's harder to directly investigate the module/class/interface of interest,
for example, the results of argument parsing here.

jl-wynen (approved):
Ok. But note the extra cases added in #133

PR #133, opened 2024-02-16
Support most callables

### LINKED ISSUE #114: Allow objects as providers
I don't see any a priori reason for only allowing functions as providers. Any callable should work. For example, the following is equivalent to a function with bound parameters (with `functools.partial`):
```python
from typing import NewType
from sciline import Pipeline

Data = NewType('Data', int)

class Provider:
    def __init__(self, param: int) -> None:
        self._param = param

    def __call__(self) -> Data:
        return Data(3 * self._param)

pl = Pipeline((Provider(2),))
```
But it raises
```
TypeError: <__main__.Provider object at 0x7fa79eb4cfd0> is not a module, class, method, or function.
```

Supporting this would

- allow for more flexibility. Such a provider can maintain internal state as long as that state does not affect the output value (e.g. caches). Or it could be used to inject objects that don't exist at the point where `Pipeline.build` gets called.
- bring Sciline closer to the OOP world that is, in my experience, more familiar to most Python programmers than the functional world.

jl-wynen:
Fixes #114

### DIFF
--- a/src/sciline/_provider.py
+++ b/src/sciline/_provider.py
@@ -123,7 +123,7 @@ class Provider:
 
     def deduce_key(self) -> Any:
         """Attempt to determine the key (return type) of the provider."""
-        if (key := get_type_hints(self._func).get('return')) is None:
+        if (key := self._arg_spec.return_) is None:
             raise ValueError(
                 f'Provider {self} lacks type-hint for return value or returns None.'
             )
@@ -165,10 +165,13 @@ class Provider:
 class ArgSpec:
     """Argument specification for a provider."""
 
-    def __init__(self, *, args: dict[str, Key], kwargs: dict[str, Key]) -> None:
+    def __init__(
+        self, *, args: dict[str, Key], kwargs: dict[str, Key], return_: Optional[Key]
+    ) -> None:
         """Build from components, use dedicated creation functions instead."""
         self._args = args
         self._kwargs = kwargs
+        self._return = return_
 
     @classmethod
     def from_function(cls, provider: ToProvider) -> ArgSpec:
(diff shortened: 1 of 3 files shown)


src/sciline/_provider.py
```diff
@@ -259,7 +274,46 @@ def _module_name(x: Any) -> str:
...
+) -> tuple[list[str], list[str], dict[str, Any]]:
+    if inspect.isfunction(func):
+        return _get_func_args_and_types(func)
+    if inspect.ismethod(func):
+        return _get_method_args_and_types(func)
+    if isinstance(func, type):
+        raise TypeError(
+            "Classes are not supported as providers because their type annotations "
+            f"cannot reliably be extracted. Got class {func}. "
+            "Consider using a free function, classmethod, or staticmethod."
```
SimonHeybrock, line 294: Is it worth mentioning that this is about classes, not class instances? While clear to us, an average user may forget to instantiate a class and not know the difference?

src/sciline/_provider.py
```diff
@@ -259,7 +274,46 @@ def _module_name(x: Any) -> str:
...
+
+def _get_method_args_and_types(
+    func: ToProvider,
+) -> tuple[list[str], list[str], dict[str, Any]]:
+    func = _unwrap_decorated(func)
+    args, kwonlyargs, hints = _get_func_args_and_types(func)
+    # Remove annotation of `self` / `cls` if present
+    hints.pop(args[0], None)
+    # Drop `self` / `cls`
+    args = args[1:]
```
SimonHeybrock, line 318: What if this is a `staticmethod`? Does this actually take the `inspect.isfunction` branch?

SimonHeybrock (commented):
Missing a test with `classmethod`?

jl-wynen:
> Missing a test with `classmethod`?

No, it's here: https://github.com/scipp/sciline/blob/1a75b95cef64887bdeeed8fd16899fcc20044192/tests/pipeline_test.py#L1495

src/sciline/_provider.py
jl-wynen: Yes. See https://github.com/scipp/sciline/blob/1a75b95cef64887bdeeed8fd16899fcc20044192/tests/pipeline_test.py#L1507

src/sciline/_provider.py
jl-wynen: Better?

SimonHeybrock (approved):

Issue #108, opened 2024-01-23
Add `__contains__`?


SimonHeybrock:
I could be useful to be able to check if a pipeline can compute a certain product, e.g.,

```python
assert WavelengthBands not in pipeline
```

jl-wynen:
We could add a named method instead of overloading `__contains__` to avoid conceptual issues.

SimonHeybrock:
I think we should add `__contains__`, returning whether a provider exists, ignoring whether its dependencies are fulfilled.

We can consider adding a named method for computability, but we should probably postpone that until we have a clearer picture about if and how we may use a graph structure inside `Pipeline`.

jl-wynen:
I don't understand. This sounds like you want to go for the option that may have conceptual issues in favour of the one where that is less of a problem with the justification that we're unsure about the interface. This seems backwards. Can yo uexplain?

SimonHeybrock:
No, I am suggesting to *not* implement the one that has conceptual issues. Instead implement a plain companion to `__setitem__` (and `insert`) and `__getitem__` that essentially just returns `tp in self._providers` (modulo complications for generic providers).

jl-wynen:
How does this differ from your original post? It is a check for whether the pipeline contains a provider that can produce the given type. This is how I understood your original idea.

SimonHeybrock:
The original suggested to return whether the pipeline can *compute* the result. This implies a transitive check whether all dependencies of the provider are fulfilled.

jl-wynen:
I see. I'm definitely against supporting this, I never even considered that you could mean this.

But the conceptual issue I was talking about is that, with your suggestion, `Pipeline` is treated either as a container of keys or as a mapping from keys to providers. This is not quite how the rest of it works. `__setitem__` only works with parameters, i.e., a subset of the mapping. And the other methods, including `insert`, `get`, `build`, `compute`, the mapping is an implementation detail.
So I'm leaning towards not using a `__contains__` method here.

SimonHeybrock:
So you are suggesting to also not implement #98 (`__getitem__`) as suggested?

> `Pipeline` is treated either as a container of keys or as a mapping from keys to providers. This is not quite how the rest of it works. `__setitem__` only works with parameters, i.e., a subset of the mapping.

I think we could easily extend `__setitem__` to also work with other providers. `insert` exists mainly to avoid having to specify types twice, but otherwise it does the same.

jl-wynen:
[...]

PR #135, opened 2024-02-19
feat: add method to get provider from type

### LINKED ISSUE #98: Provide access to code of providers via pipeline object
In many situations, it is handy to have access to the source of certain `providers` within a given pipeline in read-only mode.
While the `visualize()` functionality is already convenient for getting an overview of the pipeline, the internals of the providers are hidden, possibly even in a multilevel module.

I propose a new method which prints the path, docstring, and source of the provider, like:
```python
pipeline = sciline.Pipeline(providers, params=params)
pipeline.prdef(CleanedData) 
```
where `CleanedData` is the key to the provider.
Under the hood, one should be able to do all the magic using the `inspect` python package.

Another maybe more fancy idea is to show the docstring of `providers` in the TaskGraph as tooltips and maybe the same for type definitions of domain types.

Best

Daniel

jokasimr:
Fixes #98 and #108

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -433,6 +433,18 @@ class Pipeline:
             )
         self._set_provider(key, Provider.parameter(param))
 
+    def __getitem__(self, tp: Union[Type[T], Item[T]]) -> Provider:
+        return self.get(tp).graph[tp]
+
+    def __contains__(self, tp: Union[Type[T], Item[T]]):
+        try:
+            self.get(tp)
+        except AmbiguousProvider:
+            return True
+        except UnsatisfiedRequirement:
+            return False
+        return True
+
     def set_param_table(self, params: ParamTable) -> None:
         """
         Set a parameter table for a row dimension.
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -436,7 +436,7 @@ class Pipeline:
     def __getitem__(self, tp: Union[Type[T], Item[T]]) -> Provider:
         return self.get(tp).graph[tp]
 
-    def __contains__(self, tp: Union[Type[T], Item[T]]):
+    def __contains__(self, tp: Union[Type[T], Item[T]]) -> bool:
         try:
             self.get(tp)
         except AmbiguousProvider:
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -434,11 +434,11 @@ class Pipeline:
         self._set_provider(key, Provider.parameter(param))
 
     def __getitem__(self, tp: Union[Type[T], Item[T]]) -> Provider:
-        return self.get(tp).graph[tp]
+        return self._get_unique_provider(tp, HandleAsBuildTimeException())
 
     def __contains__(self, tp: Union[Type[T], Item[T]]) -> bool:
         try:
-            self.get(tp)
+            self[tp]
         except AmbiguousProvider:
             return True
         except UnsatisfiedRequirement:
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -433,17 +433,25 @@ class Pipeline:
             )
         self._set_provider(key, Provider.parameter(param))
 
-    def __getitem__(self, tp: Union[Type[T], Item[T]]) -> Provider:
-        return self._get_unique_provider(tp, HandleAsBuildTimeException())[0]
+    def get_provider(
+        self, tp: Union[Type[T], Item[T]]
+    ) -> Union[Callable[[Any, ...], T], T]:
+        '''Get the provider that produces the type. If the type is provided by a
+        parameter the method returns the value of the parameter.'''
+        provider = self._get_unique_provider(tp, HandleAsBuildTimeException())[0]
+        if provider.kind == 'function':
+            return provider.func
+        return provider.func()
 
-    def __contains__(self, tp: Union[Type[T], Item[T]]) -> bool:
+    def has_provider(self, tp: Union[Type[T], Item[T]]) -> bool:
+        '''Determines if the pipeline has a provider that produces the type.'''
         try:
-            self[tp]
+            self.get_provider(tp)
+            return True
         except AmbiguousProvider:
             return True
         except UnsatisfiedRequirement:
             return False
-        return True
 
     def set_param_table(self, params: ParamTable) -> None:
         """
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -539,12 +539,20 @@ class Pipeline:
         handler = handler or HandleAsBuildTimeException()
         explanation: List[str] = []
 
-        matches = [
-            (provider, bound)
-            for ptype, provider in self._providers.items()
-            if (bound := _find_bounds_to_make_compatible_type_tuple((tp,), (ptype,)))
-            is not None
-        ]
+        if tp in self._providers:
+            # Optimization to quickly find non-generic providers
+            matches = [(self._providers[tp], {})]
+        else:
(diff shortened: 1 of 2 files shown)


jokasimr:
I think that now after [adding the type matching logic](https://github.com/scipp/sciline/commit/3dc201780d4f078d04fdf27fe7e2882c6d5c5c11) we don't need to treat subproviders separately anymore.

That would simplify adding the `__getitem__` and `__contains__` methods. So therefore I'm considering making that refactoring here, what do you think @SimonHeybrock ?

src/sciline/pipeline.py
```diff
@@ -532,72 +531,59 @@ def _set_provider(
...
-                        ''.join(
-                            map(
-                                str,
+
+        matches = [
+            (provider, bound)
+            for ptype, provider in self._providers.items()
+            if (bound := _find_bounds_to_make_compatible_type_tuple((tp,), (ptype,)))
+            is not None
+        ]
```
SimonHeybrock, line 547: This replaces a constant-time lookup by a linear-time iteration. And this is called for every requirement, recursively. That is we can the complexity of `Pipeline.get` from something like `O(N)` to `O(N^2)`.

 I do not know how large pipelines will get, but this may be severe problem we should not simply ignore.

Can you somehow keep the rest of your refactor, but still have a constant-time lookup by default, falling back to iteration only if there is no match?

jokasimr: You're absolutely correct. We could group the providers by their `origin` type ahead of the iteration to achieve the same effect.

However, I tested this for the esssans workflow using
```python
%%timeit
iofq = pipeline.get(BackgroundSubtractedIofQ)
```
and the result is that the time goes from 2ms to 10ms.

I'd say that's not particularly significant and right now I don't think it is worth complicating the code for.
We can always add this in the future if it becomes an issue.

SimonHeybrock: I think I disagree. The ESSsans workflow is "tiny", and I am surprised that even there this takes 5x the time! Since the effect is **quadratic** I do not think we should ignore this.

Why do yousuggest we would have to pre-group? Can't we look for an exact match, simply calling `self._providers.get(tp)` and use that if found?

jokasimr: > Can't we look for an exact match, simply calling self._providers.get(tp) and use that if found?

Yeah sure we can do that.

I don't agree 5x is dramatic. Consider that if the effect was truly `O(n^2)` we could expect a slowdown proportional to the number of providers in the pipeline, that is, >50.

SimonHeybrock: > > Can't we look for an exact match, simply calling self._providers.get(tp) and use that if found?
> 
> Yeah sure we can do that.
> 
> I don't agree 5x is dramatic. Consider that if the effect was truly `O(n^2)` we could expect a slowdown proportional to the number of providers in the pipeline, that is, >50.

I would not expect 50x, since the code also does *other* things, apart from provider lookup.

But to clarify:

- Switching to an algorithm with much worse scaling (here `O(N^2)` instead of`O(N)` should never be done lightheartedly and without great care, *in particular* if we are not in *full* control over all the contexts where the algorithm is used (see next point).
- Sciline is a *generic* package that is meant to be useful well beyond the context of the neutron-scattering data-reduction workflows that we are writing. Therefore, the benchmark you ran above is not a sufficient data point (or rather, it shows that we should *not* make the change).

src/sciline/pipeline.py
```diff
@@ -434,6 +433,18 @@ def __setitem__(self, key: Type[T], param: T) -> None:
             )
         self._set_provider(key, Provider.parameter(param))
 
+    def __getitem__(self, tp: Union[Type[T], Item[T]]) -> Provider:
```
jl-wynen, line 436: This makes `Provider` part of the public interface. So can you import it as well as ArgSpec, ProviderLocation, ProviderKind, ToProvider in the top level __init__.py? (ToProvider maybe should be exposed in typing instead)

src/sciline/pipeline.py
jl-wynen, line 1: About the conceptual issues, I was wrong stating that this makes Pipeline a Mapping. It does not because we would have to also implement `__iter__` and `__len__`. And actually, since there is a `__setitem__`, if anything, it should be `MutableMapping`, thus we would also need `__delitem__`.
However, Pipeline implements part of the interface (and also the interface of `Sequence`). I'm not entirely sure how it fits in.

More concretely, implementing `__getitem__` makes Python think that pipelines are iterable with integer indices:
```python
import sciline as sl

def to_string(i: int) -> str:
    return str(i)

pl = sl.Pipeline([to_string], params={int: 3})
list(pl)
```
raises
```
sciline.handler.UnsatisfiedRequirement: ('No provider found for type', 0)
```
This not great. So you would also need to implement `__iter__`. And since you would be going for a dict-like interface, you would also need `keys`, `values`, and `items`.

And that leaves us with the asymmetry between `__getitem__` and `__setitem__`. For generic providers, the former currently returns different values and accepts different keys than the latter. Even if `__setitem__` were extended to support providers on top of parameters.
Also, what about parameter tables or sentinels? Would `__getitem__` return those special, internal providers? Or how would it represent them?

With the proposed semantics, the new functions allow checking what a pipeline can produce and how. So it does not allow treating Pipeline like a container. So named methods seem more appropriate to me.

SimonHeybrock: > I don't know, what do you think? I'm thinking that it's likely that a user wants to access the parameter tables that a pipeline uses, but they should preferably be returned in the same form that they were inserted, i.e. as sl.ParamTable.

Just like the discussion about using a graph data structure inside `Pipeline`, this shows that the implicit way in which we handle param-tables is problematic. We may need to re-think that (come up with a more explicit interface, or way the express the nesting?).

src/sciline/pipeline.py
```diff
@@ -434,6 +433,26 @@ def __setitem__(self, key: Type[T], param: T) -> None:
...
         self._set_provider(key, Provider.parameter(param))
 
+    def get_provider(
+        self, tp: Union[Type[T], Item[T]]
+    ) -> Union[Callable[[Any, ...], T], T]:
+        '''Get the provider that produces the type. If the type is provided by a
+        parameter the method returns the value of the parameter.'''
+        provider = self._get_unique_provider(tp, HandleAsBuildTimeException())[0]
+        if provider.kind == 'function':
+            return provider.func
```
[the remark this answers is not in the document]
jokasimr: Because I agreed with the concern that the `Provider` class isn't currently exposed, and returning a `Provider` here would expose it. I think just returning the same type that was provided by the user keeps things simpler.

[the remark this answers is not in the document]
jl-wynen: I'm not concerned about exposing it. This can be useful in its own right. 

Your solution makes it difficult to distinguish between providers and parameters. You can't check the type of the return value of `get_provider` because parameters can be callable and providers can now be any callable object, not just functions. So you would have to compare the returned type to `tp`.

src/sciline/pipeline.py
```diff
@@ -434,6 +433,26 @@ def __setitem__(self, key: Type[T], param: T) -> None:
...
+        return provider.func()
```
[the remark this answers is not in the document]
jl-wynen: Well, users might request `Series`. But that, along with the other kinds, may be subject to change if param tables are removed.

src/sciline/pipeline.py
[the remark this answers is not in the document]
jokasimr: My only concern about expose `Provider` is that it would increase the api surface area. I agree that it's a problem that you can't generally determine if the returned provider is a parameter or a function.

PR #150, opened 2024-04-02
Fix recursion error when using multiple Pipeline.set_param_series

SimonHeybrock:


tests/pipeline_with_param_table_test.py
```diff
@@ -719,3 +719,23 @@ def to_str(x: int) -> str:
     pl = sl.Pipeline((to_str,))
     pl.set_param_series(int, ints)
     assert pl.compute(sl.Series[int, str]) == sl.Series(int, {1: '1', 2: '2', 3: '3'})
+
+
+def test_multiple_param_series_can_be_broadcast() -> None:
```
nvaytet, line 724: Did you also try putting the beam center finder provider back into the loki direct beam workflow with multiple files to see if it fixes it?

nvaytet: I can also try it if you prefer.

SimonHeybrock: I did not try, no. Go ahead, since you know which pieces to touch!

nvaytet (approved):

PR #152, opened 2024-04-03
Fix blue cluster boxes in task graph visualization

### LINKED ISSUE #151: Grouping in blue boxes is gone from task graph visualization
Seems to have happened after #124 was merged.

nvaytet:
Make sure we still have clusters in the visualization when using generics.

Fixes #151

### DIFF
--- a/src/sciline/visualize.py
+++ b/src/sciline/visualize.py
@@ -87,7 +87,7 @@ def _to_subgraphs(graph: FormattedGraph) -> Dict[str, FormattedGraph]:
         if kind == 'series':
             # Example: Series[RowId, Material[Country]] -> RowId, Material[Country]
             return name.partition('[')[-1].rpartition(']')[0]
-        return name
+        return name.split('[')[0]
 
     subgraphs: Dict[str, FormattedGraph] = {}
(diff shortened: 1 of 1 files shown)


src/sciline/visualize.py
```diff
@@ -87,7 +87,7 @@ def get_subgraph_name(name: str, kind: str) -> str:
         if kind == 'series':
             # Example: Series[RowId, Material[Country]] -> RowId, Material[Country]
             return name.partition('[')[-1].rpartition(']')[0]
-        return name
+        return name.split('[')[0]
```
SimonHeybrock, line 90: Where/how was this actually broken?

SimonHeybrock: What about this discussion, where you pointed out something related? https://github.com/scipp/sciline/pull/124/files#r1489495340

nvaytet: It never made clusters, it always returned `RawData[Sample]` and `RawData[Background]` instead of `RawData` for both which would then make a cluster, I think?

SimonHeybrock (approved):

Issue #145, opened 2024-03-13
Python 3.12 type aliases don't work with parameters


jl-wynen:
The code uses `isinstance(param, key)` which doesn't work when `key` is a `TypeAliasType`. Might be fixed by #144

SimonHeybrock:
Note also #81.

SimonHeybrock:
The `isinstance` checks have been removed in the latest release so this should be done and we can implement #81.

Issue #144, opened 2024-03-13
More flexible parameter type handling


jl-wynen:
Currently, Pipeline is very strict about the types of parameters. The concrete value must exactly match the parameter key (type). It is good to have some validation but we should be more flexible.

I think we should convert the given values to the parameter key. This would allow, e.g.,
```python
params = {
    numbers.Real: 3.14,
    Path: "some/path",
    sc.DataGroup: {'a': 'dict'},
}
```
As mentioned in point 3 from https://github.com/scipp/sciline/issues/139, we have #75 but that is not sufficient for the above.

SimonHeybrock:
I think I generally agree, but would like to add a negative example, so we can base our decision also on that:

```python
A = NewType('A', float)
B = NewType('B', float)
my_param = A(1.5)
params = {
    A: 3.14,
    B: "my_param,  # converts an A to an B?
}
```

Can this be prevented, while still supporting what is suggested above?

jl-wynen:
> Can this be prevented, while still supporting what is suggested above?

I think it could be by comparing the type to *all* parameters and rejecting it if there is any match. So it's not exactly cheap.

Alternatively, we could check whether the value is an instance of `NewType` or `TypeAliasType` and not perform conversions in that case. But that would somewhat limit the use of type aliases that weren't meant as Sciline domain types. But that seems like a rare edge case.

SimonHeybrock:
I think we can consider this fixed by the removal if the `isinstance` checks?

Issue #140, opened 2024-03-04
Supporting different file handle types is too difficult


jl-wynen:
In the context of https://github.com/scipp/ess/issues/161 we considered supporting opening files once and passing the file handle to the various loaders. Supporting this is tricky because there are many different file handle types:
```python
open('..', 'rb', buffering=True) -> io.BufferedReader
open('..', 'rb', buffering=False) -> io.FileIO
io.BytesIO(...) -> io.BytesIO
```

So supporting these with the current system requires
```python
File = NewType('File', io.FileIO)
BufferedFile = NewType('BufferedFile', io.BufferedReader)
Buffer = NewType('Buffer', io.BytesIO)

def load(file: Union[File, BufferedFile, Buffer]) -> X: ...

with open('..', 'rb') as f:
   sciline.Pipeline([load], params={BufferedFile: BufferedFile(f)
```
where the param key needs to be adjusted for the concrete type of file handle. This is bad UX and Python has a workaround. For normal functions, we would annotate them as
```python
def ideal_load(file: typing.BinaryIO) -> X: ...
```
but that is not possible because `Pipeline` requires an exact type match for parameters, i.e.
```python
File = NewType('File', typing.BinaryIO)
with open('..', 'rb') as f:
   sciline.Pipeline([load], params={File: File(f)
```
fails.

This gets exasperated when `load` needs to also support a file path and maybe an open `h5py.File` or `scippnexus.File`.

I don't know a good solution at this point. But automatic type conversion in parameters might help (as proposed as part of scipp/essreduce#139).

SimonHeybrock:
Maybe the type check in `__setitem__` is not a good idea after all? Or would it be possible to relax the check to a base class or protocol?

I have also seen trouble with Python 3.12 type aliases in relation to that, but I need to try that again as I do not remember the details.

YooSunYoung:
> Maybe the type check in __setitem__ is not a good idea after all? Or would it be possible to relax the check to a base class or protocol?

I think checking if the type if a subclass of the desired type should be enough.
It's also compatible with static type checkers.
Like... 
```python
NewInt = NewType("NewInt", int)

a: int = NewInt(1)  # pass the static type check
b: NewInt = 1  # does not pass the static type check
```

jl-wynen:
[...]

PR #153, opened 2024-04-11
Add ADR 0001: Remove isinstance checks when setting parameters

SimonHeybrock:
Please comment!

### DIFF
--- a/docs/developer/adr/0001-remove-isinstance-checks-when-setting-parameters.md
+++ b/docs/developer/adr/0001-remove-isinstance-checks-when-setting-parameters.md
@@ -1,0 +1,43 @@
+# ADR 0001: Remove isinstance checks when setting parameters
+
+- Status: proposed
+- Deciders:
+- Date: 2024-04-11
+
+## Context
+
+Sciline builds a data dependency graph based on type hints of callables.
+Dependencies can be fulfilled by setting values (instances of classes) as so called *parameters*.
+In an attempt to extend the correctness guarantees of the dependency graph, Sciline's `__setitem__` checks if the value is instance of the key (a type) when setting a parameter.
+
+This has led to a number of problems.
+For example, supporting different file handles types is too difficult [#140](https://github.com/scipp/sciline/issues/140),
+parameter type handling is too inflexible in general [#144](https://github.com/scipp/sciline/issues/144),
+and the mechanism is broken with Python 3.12 type aliases [#145](https://github.com/scipp/sciline/issues/145).
+In short, the mechanism gets in the way of the user, since it causes false positives.
+
+Considering the bigger picture, we can think of this mechanism as a poor man's form of *validation*.
+Validation of input parameters is very important when running workflows, but it should be done in a more explicit way.
+Validating the type is only a fraction of what we want to do when validating parameters.
+Therefore, we should remove this mechanism and replace it with a more general validation mechanism.
+The more general validation mechanism can be considered out of scope for Sciline, and should be implemented in the user code or using other common libraries such as `pydantic`.
+
+Finally, we can think of this mechanism as a form of runtime type checking.
+This is not the intended scope of Sciline.
+
+## Decision
+
+- Remove the mechanism that checks if a value is an instance of the key when setting it as a parameter.
+- Encourage users to validate inputs in providers, which can also be tested in unit tests without setting up the full workflow.
+- Encourage users to use a more general parameter validation mechanism using other libraries.
+
+## Consequences
+
+### Positive
+
+- The mechanism will no longer get in the way of the user.
+- The code will be simplified slightly.
+
+### Negative
+
+- The correctness guarantees will be slightly diminished.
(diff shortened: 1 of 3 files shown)


docs/developer/adr/0001-remove-isinstance-checks-when-setting-parameters.md
```diff
@@ -0,0 +1,43 @@
...
+The more general validation mechanism can be considered out of scope for Sciline, and should be implemented in the user code or using other common libraries such as `pydantic`.
+
+Finally, we can think of this mechanism as a form of runtime type checking.
+This is not the intended scope of Sciline.
+
+## Decision
+
+- Remove the mechanism that checks if a value is an instance of the key when setting it as a parameter.
+- Encourage users to validate inputs in providers, which can also be tested in unit tests without setting up the full workflow.
+- Encourage users to use a more general parameter validation mechanism using other libraries.
```
nvaytet, line 32: Maybe we can also provide a small framework to help users carry out validation?

nvaytet (commented):
Looking at the reasons you list, I think removing it is the right thing to do, so :+1: from my side.

jokasimr:
:+1:

docs/developer/adr/0001-remove-isinstance-checks-when-setting-parameters.md
jl-wynen, line 1: I'm not sure this is a good idea. It would not solve #140 and #144. E.g., this
```python
def foo(p: Path) -> str:
  return p.read_text()

pl = Pipeline([foo], params={Path: "my/file"})
tg = pl.get(str)
```
would work but it would fail during computation.

As I see it, this introduces duck-typing for functions that are annotated with strict type annotations and thus violate the assumptions of the author of the function.
To give another example, a function annotated to take a `DataGroup` may be able to work with any `Mapping`. But that would be coincidental and an implementation detail because the function is allowed to use the full interface of `DataGroup`.

So I think this proposal only solves #145.

You can argue that at runtime, we strictly speaking always have duck-typing. But that can be checked by a type checker. What you  are proposing is uncheckable, as far as I can tell.

jokasimr: I don't think it violates the assumptions of the author. Unless we explicitly promise to validate types, users should assume the same level of type checking as anywhere else in python land (that is, duck typing).

I think you're right that setting parameter values would be difficult to check by a type checker.

SimonHeybrock: > As I see it, this introduces duck-typing for functions that are annotated with strict type annotations and thus violate the assumptions of the author of the function.

Not sure I understand the argument. Don't you have the same problem with providers that lie about their return type? Is your concern that providers can be checked with `mypy` but `Pipeline.__setitem__` cannot?

Regarding `DataGroup`, in general I would argue that the use of domain types (type aliases) is a way of encoding a potentially very complex contract into a brief name. Does this absolve the consumer from checking relevant parts of the contract? Is it not allowed to fail if the contract is violated? I would see the type as part of the contract.

Not sure why you think #144 is not solved. We can argue about #140, but wait for ADR-0002 where I will propose to remove special handling of `Optional` and `Union`.

docs/developer/adr/0001-remove-isinstance-checks-when-setting-parameters.md
```diff
@@ -23,13 +23,38 @@ Therefore, we should remove this mechanism and replace it with a more general va
...
+
+```python
+pipeline[QBins] = sc.linspace(...)
+pipeline[QBins] = 1000  # error in current implementation
+pipeline[QBins] = sc.linspace(..., unit='m')  # no error, but wrong unit
+```
+
+Checking the type catches the first error, but not the second.
+Paradoxically, setting an integer would often be a valid operation in the example, since `scipp.hist` can handle this case, whereas the wrong unit would not be valid.
+This may indicate that defining `QBins` as an alias of `scipp.Variable` is actually an anti-pattern.
```
jl-wynen, line 41:
```suggestion
This may indicate that defining `QBins` as an alias of `scipp.Variable` is not precise enough.
```
It's a too specific case to call it a pattern.

docs/developer/adr/0001-remove-isinstance-checks-when-setting-parameters.md
```diff
@@ -23,13 +23,38 @@ Therefore, we should remove this mechanism and replace it with a more general va
 The more general validation mechanism can be considered out of scope for Sciline, and should be implemented in the user code or using other common libraries such as `pydantic`.
 
 Finally, we can think of this mechanism as a form of runtime type checking.
-This is not the intended scope of Sciline.
+We should ask ourselves if this is the intended scope of Sciline.
+If it is, shouldn't we also check that each provider actually returns the correct type?
```
jl-wynen, line 27: Can you reformulate to turn it into a statement that this is not the scope of sciline?

SimonHeybrock: I guess I should put that statement into the "Decision" section?

jl-wynen: Yes

jl-wynen (approved):

PR #157, opened 2024-04-15
Simplify remove subproviders

jokasimr:
A small simplification that removes the sub-providers concept.
I think that makes the code a bit easier to understand.

This change was previously part of #135 but that will probably not be merged, so I'm moving it here because it's unrelated and useful independent of #135.

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -543,7 +543,7 @@ class Pipeline:
 
         if provider := self._providers.get(tp):
             # Optimization to quickly find non-generic providers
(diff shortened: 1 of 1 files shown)


SimonHeybrock:
Is this needed for anything right now? The re-implementation is on its way, so unless this is required I would suggest to leave `Pipeline` unchanged for now.

jokasimr:
> Is this needed for anything right now? The re-implementation is on its way, so unless this is required I would suggest to leave Pipeline unchanged for now.

I thought it might simplify the refactoring, that's why I pushed it now

PR #158, opened 2024-04-15
Implement ADR 0002 (remove special handling of Optional and Union)

SimonHeybrock:
While `Pipeline` will be re-implemented, I nevertheless made this change, since it allows for updating units tests with new expected behavior, which will make the re-implementation easier/smaller (will pass the same unit tests), decoupling the behavior change from the re-implementation.

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 from collections import defaultdict
 from collections.abc import Iterable
 from itertools import chain
+from types import UnionType
 from typing import (
     Any,
     Callable,
@@ -722,11 +723,11 @@ class Pipeline:
         ...
 
     @overload
-    def compute(self, tp: object, **kwargs: Any) -> Any:
+    def compute(self, tp: UnionType, **kwargs: Any) -> Any:
         ...
 
     def compute(
-        self, tp: type | Iterable[type] | Item[T] | object, **kwargs: Any
+        self, tp: type | Iterable[type] | Item[T] | UnionType, **kwargs: Any
     ) -> Any:
         """
         Compute result for the given keys.
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -40,7 +40,7 @@ from .handler import (
 from .param_table import ParamTable
 from .scheduler import Scheduler
 from .series import Series
-from .typing import Graph, Item, Key, Label, get_optional, get_union
+from .typing import Graph, Item, Key, Label
 
 T = TypeVar('T')
 KeyType = TypeVar('KeyType', bound=Key)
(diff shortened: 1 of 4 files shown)


src/sciline/pipeline.py
```diff
@@ -149,7 +149,7 @@ def _find_nodes_in_paths(graph: Graph, end: Key) -> List[Key]:
 
 
 def _is_multiple_keys(
-    keys: type | Iterable[type] | Item[T],
+    keys: type | Iterable[type] | Item[T] | object,
```
jl-wynen, line 152: Isn't this equivalent to `keys: Any` now? Why did you add `object`?

src/sciline/pipeline.py
```diff
@@ -793,7 +752,13 @@ def compute(self, tp: Iterable[Type[T]], **kwargs: Any) -> Dict[Type[T], T]:
     def compute(self, tp: Item[T], **kwargs: Any) -> T:
         ...
 
-    def compute(self, tp: type | Iterable[type] | Item[T], **kwargs: Any) -> Any:
+    @overload
+    def compute(self, tp: object, **kwargs: Any) -> Any:
```
jl-wynen, line 756: Similar to above, this is equivalent to 
```suggestion
    def compute(self, tp: Any, **kwargs: Any) -> Any:
```
if I'm not mistaken.
Does mypy use the specialised overloads where possible or does it always fall back to `Any`?

tests/pipeline_with_optional_test.py
```diff
@@ -1,229 +1,100 @@
...
         return f'{x or 123}'
 
     pipeline = sl.Pipeline([use_optional], params={int: 1})
-    assert pipeline.compute(str, scheduler=scheduler) == '1'
+    with pytest.raises(sl.UnsatisfiedRequirement):
+        pipeline.get(str)
 
 
-def test_union_with_none_can_be_used_instead_of_Optional() -> None:
+def test_Union_and_optional_are_distinct() -> None:
```
jl-wynen, line 67: I don't think `Union[T, None]` and `Optional[T]` should be treated differently.
But this test doesn't actually seem to have anything to do with `Optional`. It just tests different argument orders.

src/sciline/pipeline.py
SimonHeybrock: Yes, I could not get it to work otherwise. I thought listing the others still serves as an expression of intent?

tests/pipeline_with_optional_test.py
SimonHeybrock: > I don't think Union[T, None] and Optional[T] should be treated differently.

Then we would need to add special-case handling.

src/sciline/pipeline.py
jl-wynen: Maybe. So did it not work with `Any`? From what I understand that is preferred over `object`.

tests/pipeline_with_optional_test.py
jl-wynen: Ah, now I think I understand what you want to test here. But there are no optionals in the test, only unions.

MridulS:
We should probably change https://github.com/scipp/sciline/blob/5a8bdd95d10c593f9cd6dabca7ee425abdc971ee/src/sciline/pipeline.py#L358 to `Optional[Dict[Any, Any]]` maybe with a comment about the weird behaviour of what is a type vs type hint?

SimonHeybrock:
I am thinking now that we should remove the `object` hints, and disable mypy warnings on a case-by-case basis where `Optional` is used. It seems either we need a fundamentally new idea how to handle `Optional`, or just avoid it entirely?

PR #159, opened 2024-04-15
Drop Python 3.9 support

SimonHeybrock:


SimonHeybrock:
@jl-wynen Any idea why the serialization tests are failing here? They were disabled for Python 3.9, but are broken now for 3.10.

SimonHeybrock:
> @jl-wynen Any idea why the serialization tests are failing here? They were disabled for Python 3.9, but are broken now for 3.10.

I fixed it, but I was wondering if this was expected.

tests/serialize/json_test.py
jl-wynen, line 1: Do you want to keep the pytest skip mark on the tests? I think it can be removed.

tests/serialize/json_test.py
```diff
@@ -98,55 +98,55 @@ def node_sort_key(node: dict[str, Any]) -> str:
         'id': '1',
         'label': 'to_string',
         'kind': 'function',
-        'function': 'json_test.to_string',
+        'function': 'tests.serialize.json_test.to_string',
```
jl-wynen, line 101: No, I didn't expect this. It is caused by `--import-mode=importlib` in pytest that was added in this PR.

SimonHeybrock: ... but is this change fine?

jl-wynen: I'd say so. This is a particular detail of how pytest works. In a normal Python process, it should always use the full module path.

jl-wynen (approved):

PR #160, opened 2024-04-15
Implement ADR 0001: Remove isinstance checks when setting params

SimonHeybrock:


jl-wynen (approved):

SimonHeybrock:
Merging into feature branch, since I want to switch to 3.10 union syntax there to fix mypy, and it would require changing the `isinstance` checks being removed here.

PR #162, opened 2024-04-17
MAINT: Error out gracefully with an unsupported scheduler

MridulS:


SimonHeybrock (approved):
👍

SimonHeybrock:
mypy is failing.

MridulS:
Yeah, runtime checks conflicts with mypy.

I don't know what's the best practice here. If there is type annotated code already should there be no unit test for isinstance checks? Or should mypy just be silenced there? mypy here will also complain about the code being unreachable as the variable `scheduler` is already type annoated as `Scheduler` so there shouldn't be a case where `scheduler` is not an instance of something that follows the `Scheduler` protocol according to the type information. But this requires the user to write their code in a type checking enabled IDE (I was doing this is in a vanilla jupyter notebook and hence tripping up on this)

SimonHeybrock:
> Yeah, runtime checks conflicts with mypy.
> 
> I don't know what's the best practice here. If there is type annotated code already should there be no unit test for isinstance checks? Or should mypy just be silenced there?

I do not know either, we have selectively ignored mypy in other places as well, so maybe go with that?

PR #164, opened 2024-04-29
Clean up: copier update, ruff happy, tox -e deps

MridulS:
One thing I noted while doing this and also looking around at other places in our codebase is that they are some comments all around about support for different python versions (some bits work with only above a specific python version). It could be helpful to formalise this a bit. I have added `FIXME: PY312` for the bits of code that should be removed once we reach py3.12 min version. Having something like this is helpful with quickly greping them.


src/sciline/pipeline.py
```diff
@@ -295,17 +296,18 @@ def _copy_node(
...
         }
         return SeriesProvider(selected.keys(), provider.row_dim, args=selected.values())
 
     def _find_grouping_node(self, index_name: Key, subgraph: Graph) -> type:
-        ends: List[type] = []
-        for key in subgraph:
-            if get_origin(key) == Series and get_args(key)[0] == index_name:
-                # Because of the succeeded get_origin we know it is a type
-                ends.append(key)  # type: ignore[arg-type]
+        ends: List[type] = [
```
jl-wynen, line 305: Still `List` with capital 'L'. Did you run the upgrader for type hints?

tests/param_table_test.py
```diff
@@ -6,22 +6,26 @@
...
+    with pytest.raises(
+        ValueError, match="Cannot create param table with zero columns and no index"
+    ):
         sl.ParamTable(row_dim=int, columns={})
 
 
 def test_raises_with_inconsistent_column_sizes() -> None:
-    with pytest.raises(ValueError):
+    with pytest.raises(
+        ValueError, match="Columns in param table must all have same size"
```
jl-wynen, line 17: Not sure if we should put the exact error message here. This means that the test needs to be updated when the message changes. (Here, in particular, there is a grammatical error...)

MridulS: I think we should definitely put them in. We use generic errors like `ValueError` and with a catch-all test we can't really be sure if the error is being raised at the correct place.

jl-wynen: I mean that I wouldn't put the exact error message but instead a pattern that checks for expected terms.

src/sciline/pipeline.py
MridulS: I did not. I thought we added pyupgrade to the ruff rules https://github.com/scipp/copier_template/blob/afbe5c758fbc9e42b4379939e833e1f746187dd7/template/pyproject.toml.jinja#L67 but I guess not?

jl-wynen: No, you would have to run it manually

SimonHeybrock (changes requested):
As I have major changes in-progress, can we postpone repo-wide reformatting until those are in? Resolving the conflicts would be a nightmare.

PR #165, opened 2024-05-08
Rewrite of Sciline's Pipeline

SimonHeybrock:
### Overview

This is a rewrite of `Pipeline`, as discussed in many places. Please see the `rewrite.ipynb` notebook for some context.

### List of changes

- `TypeVar`s need to be constrained. The new mechanism eagerly inserts duplicate nodes based on this.
- Mechanism to favor most-specialized provider is removed. The last provider being inserted is the one that is used. This may be a pitfall, but otherwise it is never clear if an `insert` should replace existing providers or not, and may even make it impossible to replace specialized providers with a generic one. The new implementation is very simple, so that is what we do.
- Param table support removed, in favor of `map` and `reduce` based on Cyclebane. Not that `groupby` is not supported yet, but we have never used this in practice.
- Provider input type hints must be unique.
- `__setitem__` and `__getitem__` can be used to compose more complex graphs, sidestepping the dependency-injection-driven assembly of the base pipeline.


src/sciline/_provider.py
```diff
@@ -170,10 +171,18 @@ def __init__(
...
+        # Duplicate type hints could be allowed in principle, but it makes structure
+        # analysis and checks more difficult and error prone. As there is likely
+        # little utility in supporting this, we disallow it.
+        if len(set(args.values()) | set(kwargs.values())) != len(args) + len(kwargs):
+            raise ValueError("Duplicate type hints found in args and/or kwargs")
         self._args = args
         self._kwargs = kwargs
         self._return = return_
 
+    def __len__(self) -> int:
```
jl-wynen, line 183: I don't think `__len__` is the best choice here. The user might wonder whether this includes the return annotation or not. (This is how `len(__annotations__)` works)

src/sciline/_provider.py
```diff
@@ -220,7 +229,10 @@ def keys(self) -> Generator[Key, None, None]:
 
     def bind_type_vars(self, bound: dict[TypeVar, Key]) -> ArgSpec:
         """Bind concrete types to TypeVars."""
-        return self.map_keys(lambda arg: _bind_free_typevars(arg, bound=bound))
+        out = self.map_keys(lambda arg: _bind_free_typevars(arg, bound=bound))
+        if self._return is not None:
+            out._return = _bind_free_typevars(self._return, bound=bound)
+        return out
```
jl-wynen, line 235: Should this be moved into `map_keys`?

tests/task_graph_test.py
```diff
@@ -31,7 +31,7 @@ def as_float(x: int) -> float:
 
 def make_task_graph() -> Graph:
     pl = sl.Pipeline([as_float], params={int: 1})
-    return pl.build(float, handler=sl.HandleAsBuildTimeException())
+    return pl.build((float,), handler=sl.HandleAsBuildTimeException())
```
jl-wynen, line 34: Is the interface of `build` and `get` different now? Why? This seems confusing.

jl-wynen (commented):
I haven't reviewed `DataGraph` yet. That will need more time.

pyproject.toml
```diff
@@ -30,6 +30,7 @@ requires-python = ">=3.10"
 # Run 'tox -e deps' after making changes here. This will update requirement files.
 # Make sure to list one dependency per line.
 dependencies = [
+  "cyclebane @ git+https://github.com/scipp/cyclebane@fix-conceptual-naming-issues",
```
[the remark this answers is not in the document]
SimonHeybrock: It still requires review.

src/sciline/pipeline.py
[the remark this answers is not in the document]
SimonHeybrock: My goal was to keep the interface of `DataGraph` simple, especially since it was not entirely clear if we wanted to keep the entire interface of `Pipeline` in its current form and shape. `Pipeline` is maximally backwards compatible (not fully due to param-table removal). It could of course be merged, but it felt more natural to keep `DataGraph` basic, with only the features a literal "data graph"  needed.

tests/task_graph_test.py
SimonHeybrock: `build` is a lower-level method with a simpler interface. The complexity of distinguishing between single and multiple keys was moved in to `get`. `build` should (as before) not be used by users.

src/sciline/_provider.py
SimonHeybrock: By the same argument, isn't what existing methods like `keys()` return also confusing?

jl-wynen: True. It should be more like `argument_keys`.

src/sciline/_provider.py
jl-wynen: See above. In Sciline, 'key' refers to the things that connect providers, i.e., domain types. So it applies to the return type as well.

But my question was in particular about whether it always makes sense to map the return key.

src/sciline/pipeline.py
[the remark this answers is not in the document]
jl-wynen: I don't feel the same way. I think my problem is that `Pipeline` is just a collection of functions that can be applied to data graphs. It doesn't add any new behaviour in the sense of having methods that work together to encapsulate state. Everything in Pipeline could be made into a free function without any effect on programs except for syntax.

src/sciline/_provider.py
SimonHeybrock: Since this is just an internal implementation detail, I suggest to leave this for now.

src/sciline/pipeline.py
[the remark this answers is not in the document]
SimonHeybrock: Remove `build` in favor of `to_networkx`. Using free function to make a `Graph`. The latter is now less exposed in the public interface. Still not perfect, but better?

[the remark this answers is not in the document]
jl-wynen: Better :+1:

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def _mapping_to_constrained(
+    type_vars: set[TypeVar],
+) -> Generator[dict[TypeVar, type], None, None]:
+    constraints = [_get_typevar_constraints(t) for t in type_vars]
+    if any(len(c) == 0 for c in constraints):
+        raise ValueError('Typevars must have constraints')
+    for combination in itertools.product(*constraints):
+        yield dict(zip(type_vars, combination))
```
jl-wynen, line 44:
```suggestion
        yield dict(zip(type_vars, combination, strict=True))
```
To pre-empt an unhappy ruff.

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+    @property
+    def _graph(self) -> nx.DiGraph:
+        return self._cbgraph.graph
+
+    def _get_clean_node(self, key: Key) -> Any:
+        """Return node ready for setting value or provider."""
+        if key is NoneType:
+            raise ValueError('Key must not be None')
+        if key in self._graph:
+            self._graph.remove_edges_from(list(self._graph.in_edges(key)))
```
jl-wynen, line 77:
```suggestion
            self._graph.remove_edges_from(self._graph.in_edges(key))
```
?

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+        Parameters
+        ----------
+        key:
+            Type to provide a value for.
+        param:
+            Concrete value to provide.
+        """
+        # This is a questionable approach: Using MyGeneric[T] as a key will actually
+        # not pass mypy [valid-type] checks. What we do on our side is ok, but the
+        # calling code is not.
```
jl-wynen, line 122: Can you elaborate why this fails in mypy?

Does this mean that every insertion of a param needs to have a `type: ignore[valid-type]` comment?

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+        if typevars := _find_all_typevars(key):
+            for bound in _mapping_to_constrained(typevars):
+                self[_bind_free_typevars(key, bound)] = value
+            return
+
+        # TODO If key is generic, should we support multi-sink case and update all?
+        # Would imply that we need the same for __getitem__.
```
jl-wynen, line 129: Turn into an issue?

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+        self._cbgraph[key] = (
+            value._cbgraph if isinstance(value, DataGraph) else _as_graph(key, value)
+        )
+
+    def __getitem__(self: T, key: Key) -> T:
```
jl-wynen, line 134: I just noticed that `__setitem__` and `__getitem__` are asymmetric:
```python
graph[key] = value
assert graph[key] != value
```
Can we avoid this somehow?

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+        return self._from_cyclebane(self._cbgraph[key])
+
+    def map(self: T, node_values: dict[Key, Any]) -> T:
```
jl-wynen, line 137: Please add docstrings to this and the other methods.

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+            return
+        # Trigger UnboundTypeVar error if any input typevars are not bound
+        provider = provider.bind_type_vars({})
+        self._get_clean_node(return_type)['provider'] = provider
+        for dep in provider.arg_spec.keys():
+            self._graph.add_edge(dep, return_type, key=dep)
+
+    def __setitem__(self, key: Key, value: DataGraph | Any) -> None:
+        """
+        Provide a concrete value for a type.
```
jl-wynen, line 111: Or a data graph.

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+        orig_keys = [edge[2].get('key', None) for edge in input_edges]
+        if (value := node.get('value', _no_value)) is not _no_value:
+            out[key] = Provider.parameter(value)
+        elif (provider := node.get('provider')) is not None:
+            new_key = dict(zip(orig_keys, input_nodes))
+            spec = provider.arg_spec.map_keys(new_key.get)
+            if len(spec) != len(input_nodes):
+                # This should be caught by __setitem__, but we check here to be safe.
+                raise ValueError("Corrupted graph")
+            # TODO also kwargs
```
jl-wynen, line 198: Please do this now or open an issue.

docs/developer/architecture-and-design/rewrite.ipynb
```diff
@@ -0,0 +1,281 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Current implementation\n",
+    "\n",
+    "- `sciline.Pipeline` is a box that can be filled with providers (a provider is callable that can compute a value) as well as values.\n",
+    "- Providers can provide generic types.\n",
+    "  The concrete types and values that such providers compute is determined *later*, when the pipeline is built, based on which instances of the generic outputs are requested (by other providers or by the user when building the pipeline).\n",
+    "- Parameter tables and a special `sciline.Series` type are supported to create tasks graphs with duplicate branches and \"reduction\" or grouping operations.\n",
```
jl-wynen, line 31:
```suggestion
    "- Parameter tables and a special `sciline.Series` type are supported to create task graphs with duplicate branches and \"reduction\" or grouping operations.\n",
```

docs/user-guide/generic-providers.ipynb
jl-wynen, line 1: Can you add a bit of text explaining that type vars must have constraints?

docs/user-guide/parameter-tables.ipynb
```diff
@@ -115,16 +111,25 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Then we can compute `Result` for each index in the parameter table:"
+    "Note how we used a node name of the pipeline as the column name in the parameter table.\n",
+    "For convenience we used a `pandas.DataFrame` to represent the table above, but the use of Pandas is entirely optional.\n",
+    "Equivalently the table could be represented as a `dict`, where each key corresponds to a column header and each value is a list of values for that column.\n",
```
jl-wynen, line 130: Can you add an example of this?

src/sciline/data_graph.py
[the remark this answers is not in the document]
SimonHeybrock: Cyclebane currently has some sub-optimal handling of graph creation after `map`-`reduce`. But I have not tested/benchmarked so far.

> But given that this may be done per update in Beamlime

I don't think we should or plan to do that?

src/sciline/data_graph.py
SimonHeybrock: I think NetworkX' `in_edges` returns a view, so that would break? @MridulS can correct me?

src/sciline/data_graph.py
SimonHeybrock: Apparently `MyGeneric[T]` is not a valid type. `MyGeneric[MyType]` is ok. So no, not every insertion needs a `type: ignore[valid-type]`, just the less common ones using a generic (not specialized).

src/sciline/data_graph.py
SimonHeybrock: We can avoid it be requiring wrapping the RHS in a `DataGraph`:

```
graph[key] = DataGraph.from_value(value)  # create single-node graph with value
```

The current behavior is a convenience feature for that.

src/sciline/data_graph.py
SimonHeybrock: #166.

src/sciline/data_graph.py
SimonHeybrock: #166.

src/sciline/data_graph.py
```diff
@@ -0,0 +1,205 @@
...
+    graph = nx.DiGraph()
+    graph.add_node(key, value=value)
+    return cb.Graph(graph)
+
+
+def _find_all_typevars(t: type | TypeVar) -> set[TypeVar]:
+    """Returns the set of all TypeVars in a type expression."""
+    if isinstance(t, TypeVar):
+        return {t}
+    return set(itertools.chain(*map(_find_all_typevars, get_args(t))))
```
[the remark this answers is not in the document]
SimonHeybrock: Can't get that to work, can you fix the suggested syntax?

src/sciline/data_graph.py
SimonHeybrock: Having trouble writing good docstrings here and in cases like `__setitem__` (choice of terminology, etc.), can we sit down together to work on these?

src/sciline/data_graph.py
SimonHeybrock: See above.

src/sciline/data_graph.py
jl-wynen: Would it be better to annotate `__setitem__` with something more lenient? Maybe even `Any`? I don't like having legitimate usages flagged as errors. That is, I would prefer false negatives over false positives.

src/sciline/data_graph.py
jl-wynen: That would be too inconvenient. Esp. when we looks at the user feedback we got. But what about `add_param` or `set_param` or something similar? Then we are explicit and it should avoid confusion around the setitem syntax.

src/sciline/data_graph.py
jl-wynen: Sure

src/sciline/data_graph.py
SimonHeybrock: We should chat about that, as I have struggled quite a bit coming up with something sensible that is not just `Any` everywhere.

src/sciline/data_graph.py
SimonHeybrock: I don't think that is convenient or obvious for users? Scientist may be familiar with `dict`, so if we can stick with that it would  be good?

jl-wynen: Maybe. I don't have evidence for either

src/sciline/data_graph.py
MridulS: Yes it's the view of the dictionary underneath, so need to create a new list object to remove the nodes.

SimonHeybrock:
Are there more comments here that cannot be done in a follow-up?

jl-wynen (approved):
I am fine with merging. But please make sure to add docstrings where missing in follow up work.

PR #170, opened 2024-06-18
Add `compute_mapped`

SimonHeybrock:
This fixes a big UX hurdle when computing results that use mapped nodes.

We will later on consider if this should be integrated more tightly with `Pipeline.compute`, for now this should serve as a solution that is 80% there an can be used to gather more insights/experience.


src/sciline/pipeline.py
jl-wynen, line 1: Did you consider how you could integrate this functionality into `compute`? E.g., check whether a given key is mapped and then do the equivalent of `compute_series`?

SimonHeybrock: Something like that, yes.

jl-wynen (commented):
Ok for a prototype. But in the long run, I would prefer merging this with `compute`. And I would also prefer not using pandas if possible because users might not use pandas either.

jl-wynen (approved):

SimonHeybrock:
> Ok for a prototype.

This is not meant to be a prototype.

> And I would also prefer not using pandas if possible because users might not use pandas either.

We could use a plain `dict`, but then we do not have index names (aka dims), so one would need to implement a custom data structure for this, including ultimately something like a multi-index (or, simpler something N-D like Scipp, but that won't work any more as soon as Cyclebane supports `groupby`). I don't think users would prefer using a custom class from an unknown library over a very well known and package in widespread use. Do you have something else in mind?

> But in the long run, I would prefer merging this with `compute`.

It is not yet clear to me that this is a good idea, so I refrained from expressing a preference at this point. `Pipeline.compute` can compute multiple keys (returning a `dict`, instead of a single value). So we would need to return a `dict` containing one or more `pandas.Series` (instead of a single `Series`). Would it be too confusing to get a `dict`, as `Series`, or a `dict` containing `Series`, depending on the args? It is starting too look like a bad interface. Maybe the methods for one target (or a single mapped target) should be separate from the one computing multiple targets?

docs/user-guide/parameter-tables.ipynb
```diff
@@ -168,7 +165,22 @@
    "source": [
     "Note the use of the `run_id` index.\n",
     "If the index axis of the DataFrame has no name then a default of `dim_0`, `dim_1`, etc. is used.\n",
-    "We can also visualize the task graph for computing the series of `Result` values:"
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "**Note**\n",
+    "\n",
+    "[compute_mapped](https://scipp.github.io/sciline/generated/functions/sciline.compute_mapped.html) depends on Pandas, which is not a dependency of Sciline and must be installed separately, e.g., using pip:\n",
```
jl-wynen, line 173: This link and the link below should be relative links.

src/sciline/pipeline.py
```diff
@@ -194,3 +199,93 @@ def bind_and_call(
...
+
+def get_mapped_node_names(
+    graph: Pipeline, key: type, index_names: Sequence[Hashable] | None = None
+) -> pandas.Series:
+    """
+    Given a graph with key depending on mapped nodes, return a series corresponding
+    mapped keys.
+
+    This is meant to be used in combination with :py:func:`Pipeline.map`.
+    If the key depends on multiple indices, the series will be a multi-index series.
```
jl-wynen, line 212: This took me a while to understand. How about this?
```suggestion
    Return all mapped keys for a base key.

    Given a graph with mapped nodes and a key to a node, this function returns
    the keys of mapped nodes that correspond to the key.
    This is meant to be used in combination with :py:func:`Pipeline.map`.
    If the key depends on multiple indices, the series will be a multi-index series.
```
And an example would go a long way to explaining the function.

src/sciline/pipeline.py
```diff
@@ -194,3 +199,93 @@ def bind_and_call(
...
+    -------
+    :
+        The series of mapped key names.
+    """
+    import pandas as pd
+    from cyclebane.graph import IndexValues, MappedNode, NodeName
+
+    candidates = [
+        node
+        for node in graph._cbgraph.graph.nodes
```
jl-wynen, line 237: I don't like that this uses a protected attribute of `graph`. Should `get_mapped_node_names` be a method of `Pipeline`?

src/sciline/pipeline.py
```diff
@@ -194,3 +199,93 @@ def bind_and_call(
     def _repr_html_(self) -> str:
         nodes = ((key, data) for key, data in self._graph.nodes.items())
         return pipeline_html_repr(nodes)
+
+
+def get_mapped_node_names(
+    graph: Pipeline, key: type, index_names: Sequence[Hashable] | None = None
```
jl-wynen, line 205: Is it a graph or a pipeline? We should be consistent with names.

src/sciline/pipeline.py
```diff
@@ -194,3 +199,93 @@ def bind_and_call(
     def _repr_html_(self) -> str:
         nodes = ((key, data) for key, data in self._graph.nodes.items())
         return pipeline_html_repr(nodes)
+
+
+def get_mapped_node_names(
```
jl-wynen, line 204: 'name' or 'key'? What is the difference?

src/sciline/pipeline.py
```diff
@@ -194,3 +199,93 @@ def bind_and_call(
...
+        node
+        for node in graph._cbgraph.graph.nodes
+        if isinstance(node, MappedNode) and node.name == key
+    ]
+    if index_names is not None:
+        candidates = [
+            node for node in candidates if set(node.indices) == set(index_names)
+        ]
+    if len(candidates) == 0:
+        raise ValueError(f"'{key}' is not a mapped node.")
```
jl-wynen, line 245: Should this check be before filtering by index names? It seems that `key` refers to a mapped node but that is filtered out by the index names.

src/sciline/pipeline.py
SimonHeybrock: This is definitely something that we need to consider for a lot of other functionality (in particular around planner graph operations) that we intend to implement. Either we need to make those properties accessible on the public interface, or potentially add a lot more methods to `Pipeline`. Right now I cannot say which is better? So unless you can clearly say which solution should be chosen, I'd like to keep this as it is for now.

jl-wynen: I don't know what operations you have in mind that need direct access to the graph. But I would say that we should expose a minimal set of primitive operations that can be composed in, e.g., methods of workflow classes.

jl-wynen (approved):

Issue #83, opened 2023-12-06
Get missing params from a pipeline


SimonHeybrock:
In #80 we added the ability to visually indicate missing params in a pipeline:
<img width="1335" alt="image" src="https://github.com/scipp/sciline/assets/12912489/8b999922-c97c-4a80-b8f1-cd7375598032">

It would be useful to extend the `ErrorHandler` so it can provide a list of missing params. This could then be used to generate a dict of params, or to generate a GUI with input fields.

PR #172, opened 2024-06-25
Expose and rename properties for accessing underlying graphs and indices

SimonHeybrock:
This will be required for a variety of functionality that we plan to add in Sciline or elsewhere. A simple example is #83:

```python
import networkx as nx
import pandas as pd

g = workflow.networkx_graph
nodes = [node for node in g.nodes() if len(g.in_edges(node)) == 0]
nodes = [node for node in nodes if 'provider' not in g.nodes[node]]

not_set = '(not set)'
df = pd.DataFrame(
    {'node': nodes, 'value': [g.nodes[node].get('value', not_set) for node in nodes]}
)
df
```
I don't know if this (without the Pandas bit) is something that should be added as a helper function in Sciline, or more specifically in our applications, e.g., in ESSreduce?

### DIFF
--- a/src/sciline/data_graph.py
+++ b/src/sciline/data_graph.py
@@ -9,6 +9,7 @@ from typing import Any, TypeVar, get_args
 
 import cyclebane as cb
 import networkx as nx
+from cyclebane.node_values import IndexName, IndexValue
 
 from ._provider import ArgSpec, Provider, ToProvider, _bind_free_typevars
 from .handler import ErrorHandler, HandleAsBuildTimeException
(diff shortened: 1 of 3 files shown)


src/sciline/data_graph.py
```diff
@@ -66,21 +67,39 @@ def __copy__(self: T) -> T:
...
+        """Names of the indices (dimensions) of the graph."""
+        return self._cbgraph.index_names
+
+    @property
+    def indices(self) -> dict[IndexName, Iterable[IndexValue]]:
+        """Names and values of the indices of the graph."""
+        return self._cbgraph.indices
+
+    @property
+    def cyclebane_graph(self) -> cb.Graph:
```
jl-wynen, line 80: Do we need this property? I don't see where it is used. And it adds one more concept to the interface. Users need to know what is the difference between all the different graphs.

jl-wynen (approved):

PR #175, opened 2024-07-11
Convert keys to support distributed scheduler

### LINKED ISSUE #163: Sciline trips up on dask.distributed
I tried running sciline with dask.distributed and it seems to trip up a code path in dask where it expects the keys in the task graph to be one of these types `bytes, int, float, str` and not the exotic ones we have.

I tried using the following code, maybe I did something wrong here?
``` python
from dask import distributed
from sciline.scheduler import DaskScheduler

client = distributed.Client()
scheduler = DaskScheduler(client.get)

pipeline.compute(Result, scheduler=client)
```

This results in a type error coming from https://github.com/dask/dask/blob/aa5a333fccd280b9535233865282019f7e3117ea/dask/core.py#L240

```
TypeError: Unexpected key type <class 'typing.NewType'> (key=__main__.Result)
```

I'm just suprised why don't we end up with these errors when we use `dask.threaded`

jl-wynen:
Fixes #163

There are currently very few tests that run with different schedulers. Of them one fails with the distributed scheduler because it uses a local provider and domain type which cannot be pickled. Along the same lines, many other tests would not work with the distributed scheduler because it would be unable to pickle or unpickle providers.

For example, `test_reduction_workflow` does not work because it uses providers defined in `tests/complex_workflow_test.py`. This module is not on `PYTHONPATH` and pickle can thus not unpickle the providers used by this workflow. This is a general limitation of dask / pickle that would require a lot of work to deal with. So it is not part of this PR.


src/sciline/_utils.py
```diff
@@ -31,23 +31,27 @@ def full_qualname(obj: Any) -> str:
...
 
 
-def key_name(key: Key | TypeVar) -> str:
+def key_name(key: Any) -> str:
     args = get_args(key)
+    if isinstance(key, UnionType):
+        return ' | '.join(map(key_name, args))
     if len(args):
         parameters = ', '.join(map(key_name, args))
         # getattr is a fallback for python < 3.10
```
SimonHeybrock, line 40: Can be removed now?

src/sciline/scheduler.py
```diff
@@ -107,3 +112,13 @@ def __repr__(self) -> str:
...
+
+
+def _to_dask_key(key: Hashable) -> str:
+    """Map a Sciline key to a dask key.
+
+    According to the docs (https://docs.dask.org/en/stable/spec.html#definitions),
+    keys in a Dask graph are not allowed to be types.
+    So this function converts Sciline keys (types) to strings.
+    """
+    return key_full_qualname(key)
```
SimonHeybrock, line 124: I wonder if we should do this in a base scheduler? But probably does not matter now, we can change later if ever needed elsewhere but dask?

... or does this matter for JSON serialization? Is there some duplication to be avoided if this moves to a lower level? Or to `TaskGraph`?

jl-wynen: As far as scheduling is concerned, the conversion could be done in `TaskGraph`. But it is lossy because it looses the actual domain type. So it may get in the way of other operations than scheduling.

It could be done in a base scheduler if you want it to create a separate intermediate graph with replaced keys. But given that it is only an issue for dask at the moment and that we don't have plans for any other scheduler (as far as I know), I would not generalise yet. We can always do that later if need be.

SimonHeybrock: What about the serialization and deserialization part? Was there any trouble with the latter, reconstructing the key types?

jl-wynen: It serializes the dask graph which only contains strings as keys. So no, this part does not seem to cause problems. Only providers are problematic.

SimonHeybrock (approved):

PR #186, opened 2024-09-25
fix: remove wrapping Optional if python version < 3.11

### LINKED ISSUE #182: Param setting not working with default values
For essimaging I wanted to make some arguments optional,

```python

import sciline as sl
from typing import NewType

OptionalInt = NewType("OptionalInt", int | None)


def foo(a: OptionalInt) -> str:
    return str(a) if a is not None else "Here is your default message."


wf = sl.Pipeline([foo], params={OptionalInt: OptionalInt(None)})
wf.compute(str)

# 'Here is your default message.'

```

But if the argument has a default value of None, it fails.

```python

import sciline as sl
from typing import NewType

OptionalInt = NewType("OptionalInt", int | None)


def foo(a: OptionalInt = None) -> str:
    return str(a) if a is not None else "Here is your default message."


wf = sl.Pipeline([foo], params={OptionalInt: OptionalInt(None)})
wf.compute(str)

# sciline.handler.UnsatisfiedRequirement: ('No provider found for type', typing.Optional[__main__.OptionalInt])

```

Is it a known limitation or a bug...?

jokasimr:
Fixes #182

### DIFF
--- a/src/sciline/_provider.py
+++ b/src/sciline/_provider.py
@@ -5,6 +5,7 @@
 from __future__ import annotations
 
 import inspect
+import sys
 from collections.abc import Callable, Generator, Hashable
 from dataclasses import dataclass
 from types import UnionType
@@ -303,6 +304,10 @@ def _get_func_args_and_types(
     func = _unwrap_decorated(func)
     hints = get_type_hints(func)
     signature = inspect.getfullargspec(func)
+    if sys.version_info[1] < 11:
+        for name, param in inspect.signature(func).parameters.items():
+            if param.default is None:
+                hints[name] = get_args(hints[name])[0]
     return signature.args, signature.kwonlyargs, hints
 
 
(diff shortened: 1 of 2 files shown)


src/sciline/_provider.py
```diff
@@ -303,6 +304,10 @@ def _get_func_args_and_types(
     func = _unwrap_decorated(func)
     hints = get_type_hints(func)
     signature = inspect.getfullargspec(func)
+    if sys.version_info[1] < 11:
+        for name, param in inspect.signature(func).parameters.items():
+            if param.default is None:
+                hints[name] = get_args(hints[name])[0]
```
SimonHeybrock, line 310: What happens if the function signature is `f(a: int | None = None)`? We don't want to remove the optional in that case!

jokasimr: Aha, I thought an additional optional was added as a wrapper also in that case, but I guess it isn't.

jokasimr: I don't know if there's a good way to fix this. Maybe better in that case to just do nothing and wait for later python versions like was suggested in the issue.

SimonHeybrock: Agree, do nothing!

PR #187, opened 2024-10-16
Add `mode` argument to `visualize` for more compact data or task graph display

### LINKED ISSUE #82: Increase graph readability when using parametrized providers
Example:
image

We have no clear idea about a solution right now.

SimonHeybrock:
Here is an before and after for ESSsans' `IofQ[SampleRun]`.
Before:
sciline-vis-before
After:
sciline-vis-after

Fixes #82.

### DIFF
--- a/src/sciline/visualize.py
+++ b/src/sciline/visualize.py
@@ -31,7 +31,7 @@ FormattedGraph = dict[str, FormattedProvider]
 def to_graphviz(
     graph: Graph,
     compact: bool = False,
-    mode: Literal['data', 'task', 'both'] = 'both',
+    mode: Literal['data', 'task', 'both'] = 'data',
     cluster_generics: bool = True,
     cluster_color: str | None = '#f0f0ff',
     **kwargs: Any,
@@ -117,7 +117,7 @@ def _add_subgraph(
     graph: FormattedGraph,
     dot: Digraph,
     subgraph: Digraph,
-    mode: Literal['data', 'task', 'both'] = 'both',
+    mode: Literal['data', 'task', 'both'],
 ) -> None:
     cluster = subgraph.name is not None
     cluster_connected = []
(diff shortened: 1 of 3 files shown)


src/sciline/visualize.py
```diff
@@ -82,28 +113,72 @@ def _to_subgraphs(graph: FormattedGraph) -> dict[str, FormattedGraph]:
...
+                                arg.name,
+                                ret_name,
+                                lhead=subgraph.name,
+                                # Thick pen to indicate multiple connections
+                                penwidth='2.0',
+                            )
+                            cluster_connected.append(arg.name)
+                    else:
+                        dot.edge(arg.name, ret_name)
         # else: Do not draw dummy providers created by Pipeline when setting instances
```
nvaytet, line 184: There is an error in the built docs (in parameter tables):
Screenshot at 2024-10-30 09-24-00

src/sciline/visualize.py
```diff
@@ -82,28 +113,72 @@ def _to_subgraphs(graph: FormattedGraph) -> dict[str, FormattedGraph]:
...
```
nvaytet, line 183: I don't think the curved arrows are always better. Comparing the new to old figure, while there are less boxes and duplication in the new figure, I find it harder to follow the arrows.

New:
Screenshot at 2024-10-30 09-25-48

Old:
Screenshot at 2024-10-30 09-26-06

src/sciline/visualize.py
```diff
@@ -82,28 +113,72 @@ def _to_subgraphs(graph: FormattedGraph) -> dict[str, FormattedGraph]:
...
```
nvaytet, line 182: Another error in the generic providers notebook
Screenshot at 2024-10-30 09-29-38

src/sciline/visualize.py
SimonHeybrock: Hmm, why are they curved now?

SimonHeybrock: Avoided some, see update.

nvaytet (approved):

Issue #177, opened 2024-08-14
Should `Pipeline.visualize` default to show the full graph if no keys are given?


SimonHeybrock:
See title. I think this would be quite useful.

jl-wynen:
Comment from #181:
> We have a number of pipelines across multiple projects now. And we compose new pipelines from them. It is currently very difficult to inspect a pipeline in its entirety to, e.g., find out what it can produce, or how it does a certain operation. It would be a great help if we could just visualize a pipeline without having to already know what it can produce or what products depend on the subgraph you are currently interested in.

jokasimr:
Fixed in https://github.com/scipp/sciline/pull/188

PR #188, opened 2024-10-28
feat: add method to list leaf nodes

### LINKED ISSUE #183: Listing all leaf types
## Context
I removed a leaf type to remove unnecessary step in a workflow and then user (Christian in this case) had no idea which type to visualize/compute.

It'll be easier for users if there is an interface in sciline as well to list all the leaf nodes,
and then it can be used for visualization without any type specified,
and for workflow widgets.

## Related Issues
- https://github.com/scipp/sciline/issues/177
- https://github.com/scipp/essreduce/issues/28

## Implementation Examples
There is an example already in `essreduce`.
https://github.com/scipp/essreduce/blob/aef4a3c1094dd15f1849aedc74e9960e1b922de3/src/ess/reduce/workflow.py#L52-L57

## TODO

 - [ ] Implement an interface that list all leaf nodes in the workflow
 - [ ] Use the interface for `visualize` without any arguments
 - [ ] Include the leaf nodes as examples in the `KeyError` or `NoProviderFound` error message.

jokasimr:
Fixes #183 #177

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -11,7 +11,7 @@ from ._provider import Provider, ToProvider
 from ._utils import key_name
 from .data_graph import DataGraph, to_task_graph
 from .display import pipeline_html_repr
-from .handler import ErrorHandler, HandleAsComputeTimeException
+from .handler import ErrorHandler, HandleAsComputeTimeException, UnsatisfiedRequirement
 from .scheduler import Scheduler
 from .task_graph import TaskGraph
 from .typing import Key
@@ -107,7 +107,7 @@ class Pipeline(DataGraph):
             Keyword arguments passed to :py:class:`graphviz.Digraph`.
         """
         if tp is None:
-            tp = self.leafs()
+            tp = self.final_result_keys()
         return self.get(tp, handler=HandleAsComputeTimeException()).visualize(**kwargs)
 
     def get(
@@ -140,7 +140,11 @@ class Pipeline(DataGraph):
             targets = tuple(keys)  # type: ignore[arg-type]
         else:
             targets = (keys,)  # type: ignore[assignment]
-        graph = to_task_graph(self, targets=targets, handler=handler)
+        try:
+            graph = to_task_graph(self, targets=targets, handler=handler)
+        except UnsatisfiedRequirement as e:
+            final_result_keys = ", ".join(map(repr, self.final_result_keys()))
+            raise type(e)(f'Did you meant one of: {final_result_keys}?') from e
         return TaskGraph(
             graph=graph,
             targets=targets if multi else keys,  # type: ignore[arg-type]
@@ -205,7 +209,7 @@ class Pipeline(DataGraph):
         nodes = ((key, data) for key, data in self.underlying_graph.nodes.items())
         return pipeline_html_repr(nodes)
 
-    def leafs(self) -> tuple[type, ...]:
+    def final_result_keys(self) -> tuple[type, ...]:
         """Returns the keys that are not inputs to any other providers."""
         sink_nodes = [
             cast(type, node)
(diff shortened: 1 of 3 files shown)


src/sciline/pipeline.py
```diff
@@ -200,6 +205,15 @@ def _repr_html_(self) -> str:
         nodes = ((key, data) for key, data in self.underlying_graph.nodes.items())
         return pipeline_html_repr(nodes)
 
+    def leafs(self) -> tuple[type, ...]:
```
SimonHeybrock, line 208: - Can we consider making this a free function? We should avoid bloating the interface of `Pipeline`.
- What is a leaf? Can we use some clearer terminology such as used in mathematics for DAGs, or by NetworkX? Sink nodes, for example

jokasimr: > What is a leaf? Can we use some clearer terminology such as used in mathematics for DAGs, or by NetworkX? Sink nodes, for example

I think `sink_nodes` is a good name, but it's slightly "leaking" the  graph abstraction below `Pipeline`. Maybe we should have something more specific to the domain, like `final_results` or `output_keys`.  

> Can we consider making this a free function? We should avoid bloating the interface of Pipeline.

I'm not so sure about this. We could make it a free function, but what's the advantage of doing so? It's tightly coupled to the `Pipeline` class. Having it be a method makes it easier for people to find it using dot-notation access.

src/sciline/pipeline.py
```diff
@@ -200,6 +209,15 @@ def _repr_html_(self) -> str:
         nodes = ((key, data) for key, data in self.underlying_graph.nodes.items())
         return pipeline_html_repr(nodes)
 
+    def final_result_keys(self) -> tuple[type, ...]:
+        """Returns the keys that are not inputs to any other providers."""
+        sink_nodes = [
+            cast(type, node)
```
SimonHeybrock, line 215: Not sure I understand. What is this `cast` doing?

jokasimr: It's just for appeasing mypy.
We promise the return type of the function is `tuple[type, ...]` because we know our keys are types.
But what we get from the `underlying_graph` is `Any`.

jokasimr: Okay took another look now and it was actually not necessary to cast in case we just specified the return as `tuple[Key, ...]` instead

SimonHeybrock: Indeed. Our keys are, in general not `type`.

src/sciline/pipeline.py
```diff
@@ -200,6 +209,13 @@ def _repr_html_(self) -> str:
         nodes = ((key, data) for key, data in self.underlying_graph.nodes.items())
         return pipeline_html_repr(nodes)
 
+    def final_result_keys(self) -> tuple[Key, ...]:
```
SimonHeybrock, line 212: Sorry about discussing names a second time, but can we think about the name once more. This seems a bit hard to remember, with the 3-part name. Have you considered `output_keys`? It would also be symmetric with the terminology "inputs" or input keys", which we might use elsewhere.

src/sciline/pipeline.py
jokasimr: We are using `type` in a lot of other places instead of `Key` though, that's what confused me.
For example https://github.com/scipp/sciline/blob/main/src/sciline/pipeline.py#L108-L111

Is that just a mistake or is there a difference between the contexts that makes `type` the right annotation there?

src/sciline/pipeline.py
jokasimr: No worries, yeah if you think `output_keys` is better we can use that instead :+1:

> Have you considered `output_keys`?

Yes it's one of the names I suggested the first time you brought up naming :laughing:

SimonHeybrock (approved):

PR #192, opened 2025-01-22
Add a method for connecting providers to different inputs

jl-wynen:
This is only a draft because implementing proper support for generics is non-trivial and I would like to get feedback before doing so. However, even in its current state, this implementation is useful.

This PR proposes a mechanism for replacing the inputs of providers with arbitrary nodes. I initially outlined the idea on Slack after our discussion about workflows on 2025-01-15. `override_inputs` allows side-stepping domain types and telling Sciline to use a different type instead.

### 'Lying' domain types

We have some pipelines where users can switch off operations. E.g., footprint correction in ESSreflectometry. Doing so keeps a `FootprintCorrectedData` node in the graph which is misleading.

Given a provider like
```python
def correct_by_footprint(raw: RawData, footprint: Footprint) -> FootprintCorrectedData:
    ...
```
and using
```python
pl = sciline.Pipeline([correct_by_footprint, ...])
pl = pl.override_input(FootprintCorrectedData, RawData)
```
effectively removes footprint correction from the graph. The resulting pipeline still contains `correct_by_footprint` and `FootprintCorrectedData` but they are no longer ancestors to any other nodes. This means that the correction won't be used by `compute`. And it won't show up in visualisations or when serialising to JSON.

### Choosing alternative algorithms

In powder diffraction, we allow the user to select between normalisation by proton charge and monitor. This can be achieved like so:
1. Create dedicated domain types `ProtonChargeNormalizedRunData` and `HistogramMonitorNormalizedRunData` and use those as return types of normalisation providers.
2. Create a type `NormalizedRunData` and use that as input type for all providers consuming normalised data.
3. Create a pipeline and add all providers for *all* normalisations.
4. Select a normalisation using
```python
pl = pl.override_input(NormalizedRunData, ProtonChargeNormalizedRunData)
```
Any providers that request `NormalizedRunData` will now get `ProtonChargeNormalizedRunData` instead.

I implemented this in [TODO](https://github.com/scipp/essdiffraction/pull/121)

### This is a one time operation.

The providers are not removed. This means that adding providers later can reintroduce the overridden type. This can partially be ameliorated by removing the overridden nodes. Or fixed full by creating an alias instead that affects all providers added later. (A bit like aliases in coord transforms)

### No type safety

We normally benefit from some level of type safety by relying on type hints. This is lost if we allow arbitrary overrides. I don't think this is a major problem given that Dask has no type safety either.

### Complex implementation

Especially for generic types. This may not be worth the effort. But as I showed above, even a simple implementation fixes the most pressing issues.

### DIFF
--- a/src/sciline/data_graph.py
+++ b/src/sciline/data_graph.py
@@ -177,6 +177,29 @@ class DataGraph:
             dot.edge(str(edge[0]), str(edge[1]), label=label)
         return dot
 
+    def override_input(self, original: Key, replacement: Key) -> DataGraph:
+        dg = self.copy()
+        dg._override_input_in_place(original, replacement)
+        return dg
+
+    def _override_input_in_place(self, original: Key, replacement: Key) -> None:
+        formal_args = getattr(replacement, '__args__', ())
+        assert getattr(original, '__args__', ()) == formal_args
+        for formal in formal_args:
+            # If `formal` is a type var, plug in all possible concrete types:
+            for actual in getattr(formal, '__constraints__', ()):
+                self._override_input_in_place(original[actual], replacement[actual])
+
+        _move_out_edges(self, original, replacement)
+
+
+def _move_out_edges(graph: DataGraph, original: Key, replacement: Key) -> None:
+    """Change all out edges from ``original`` to start from ``replacement``."""
+    g = graph.underlying_graph
+    to_replace = [edge for edge in g.edges(data=True) if edge[0] == original]
+    g.remove_edges_from(to_replace)
+    g.add_edges_from((replacement, *edge[1:]) for edge in to_replace)
+
 
 _no_value = object()
 
--- a/tests/pipeline_test.py
+++ b/tests/pipeline_test.py
@@ -1440,3 +1440,256 @@ def test_output_keys_in_not_found_error_message(get_method: str) -> None:
         getattr(pl, get_method)(int)
     for key in pl.output_keys():
         assert key_name(key) in info.value.args[0]
+
+
+def test_override_input_disconnects_from_existing_parent() -> None:
+    def foo(x: int) -> str:
+        return str(x)
+
+    pipeline = sl.Pipeline([foo], params={int: 3})
+    new = pipeline.override_input(int, float)
+
+    with pytest.raises(sl.UnsatisfiedRequirement, match="No provider.*float"):
+        new.compute(str)
+
+
+def test_override_input_disconnects_from_existing_parent_generic() -> None:
+    T1 = NewType('T1', int)
+    T2 = NewType('T2', int)
+    T = TypeVar('T', T1, T2)
+
+    class A(sl.Scope[T, int], int): ...
+
+    class B(sl.Scope[T, int], int): ...
+
+    class C(sl.Scope[T, int], int): ...
+
+    def foo(a: A[T]) -> B[T]:
+        return B[T](a + 2)
+
+    pipeline = sl.Pipeline([foo], params={A[T1]: A[T1](3), A[T2]: A[T2](4)})
+    new = pipeline.override_input(A[T], C[T])
+
+    with pytest.raises(sl.UnsatisfiedRequirement, match="No provider.*C"):
+        new.compute(B[T1])
+
+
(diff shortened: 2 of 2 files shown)


SimonHeybrock:
> ### 'Lying' domain types
> 
> We have some pipelines where users can switch off operations. E.g., footprint correction in ESSreflectometry. Doing so keeps a `FootprintCorrectedData` node in the graph which is misleading.
> 
> Given a provider like
> 
> ```python
> def correct_by_footprint(raw: RawData, footprint: Footprint) -> FootprintCorrectedData:
>     ...
> ```
> 
> and using
> 
> ```python
> pl = sciline.Pipeline([correct_by_footprint, ...])
> pl = pl.override_input(FootprintCorrectedData, RawData)
> ```
> 
> effectively removes footprint correction from the graph. The resulting pipeline still contains `correct_by_footprint` and `FootprintCorrectedData` but they are no longer ancestors to any other nodes. This means that the correction won't be used by `compute`. And it won't show up in visualisations or when serialising to JSON.

I need some more time to wrap my head around this, but I feel this mechanism is backwards. What you don't show above is the follow-up, namely a provider
```python
def next_correction(da: FootprintCorrectedData) -> MoreCorrectedData: ...
```
There are two cases:
1. The type hint is correct and the provider really needs `FootprintCorrectedData`.
   - Adding an alias with `override_input` would be incorrect.
2. The type hint is a lie, since this works also with data that does not have the previous correction.
   - Adding an alias would be "correct", in the sense that it covers up the lying type hint.

Taken together, the mechanism is thus either incorrect (case 1), or we have covered up one lie with a second lie (case 2). All this also only works because the types are all aliases for the same underlying data type. While I have no concrete suggestion for now, I feel it would be better to look into how the problems in the underlying code (definition of `next_correction`) can be addressed.

Maybe the root cause of our trouble is that what the functions (providers) actually want is to define a certain input contract. For lack of a better mechanism we have done so using domain types, but this is a "one dimensional" approach and too entangled into how Sciline builds the graphs. Is there a better way to define input contracts and related them to a graph structure that does not have this problem?

jl-wynen:
> I need some more time to wrap my head around this, but I feel this mechanism is backwards. What you don't show above is the follow-up, namely a provider
> 
> ```python
> def next_correction(da: FootprintCorrectedData) -> MoreCorrectedData: ...
> ```
> 
> There are two cases:
> 
>     1. The type hint is correct and the provider really needs `FootprintCorrectedData`.
>        
>        * Adding an alias with `override_input` would be incorrect.
> 
>     2. The type hint is a lie, since this works also with data that does not have the previous correction.
>        
>        * Adding an alias would be "correct", in the sense that it covers up the lying type hint. But the type hint is still wrong in the Python module, including in unit tests that to not operate on the Sciline level.
> 
> 
> Taken together, the mechanism is thus either incorrect (case 1), or we have covered up one lie with a second lie (case 2). All this also only works because the types are all aliases for the same underlying data type. While I have no concrete suggestion for now, I feel it would be better to look into how the problems in the underlying code (definition of `next_correction`) can be addressed.

I agree that this proposal can't work in all cases. It is a global change of the graph. So it can't deal with local changes that you described. When it comes to skipping steps, I see two distinct cases:

A user wants to experiment with **what would happen if a certain step was not performed**. In this case, quickly rewriting the graph is convenient. But it does not matter that the final result may be wrong. It was never intended to skip the step. And regardless of whether we skip it globally or only for part of the graph, the final result will be wrong to some degree.

**There is an optional step**. In this case, we need to build the pipeline around this fact. One way to do this is to turn the types around. Instead of using `override_input` to disable a step, we can use it to enable a step. This is more akin to the example in diffraction. That is, we use
```python
def fn_that_can_use_any_data(da: RawData) -> Processed:  ...
def fn_that_needs_footprint_corrected(da: FootprintCorrectedData) -> MoreCorrectedData: ...
```
and
```python
pl = pl.override_input(RawData, FootprintCorrectedData)
```
This would insert the footprint correction before any use of `RawData`. And `fn_that_needs_footprint_corrected` would still get the data it expects with or without using `override_inputs`.

jl-wynen:
Closing as there seems to be no interest in pursuing this further.

PR #193, opened 2025-02-18
Copier update + tox -e deps

jokasimr:
Two odd things happened when updating:

* The linter complained about a too long line in `requirements/make_base.py`
* The linter complained about what seems to be an error in the ci workflow:
```
.github/workflows/ci.yml:47:27: property "python-version" is not defined in object type {os: string; python: {tox-env: string; version: string}} [expression]
   |
47 |       python-version: ${{ matrix.python-version }}
   |                           ^~~~~~~~~~~~~~~~~~~~~
.github/workflows/ci.yml:48:20: property "tox-env" is not defined in object type {os: string; python: {tox-env: string; version: string}} [expression]
   |
48 |       tox-env: ${{ matrix.tox-env }}
```


jokasimr:
Why are the remaining checks not running? :thinking:

MridulS:
Yeah `Tests (ubuntu-22.04, *) / *` needs to be `Tests (ubuntu-24.04, *) / *`

MridulS:
@jl-wynen Could you update the text in required tests for PRs? Looks like only the owners can do that.

Issue #26, opened 2023-07-26
Progress reporting (documentation)


SimonHeybrock:
One could implement some hooks for some other progress-bar libraries out there. However, for now this seems overkill. We should instead recommend using the progress-reporting that comes with `dask`.

Add a documentation page:

- Providing relevant `dask` references (there are multiple options, depending on which scheduler is used).
- Give an example for the simplest / typical case.

PR #194, opened 2025-03-13
Add progress reporters

jl-wynen:
This PR adds tools for showing the progress of a `Pipeline.compute` call. For now, I only implemented support for rich. But other libraries like tqdm could be added.

See also #26 which suggests to stick with Dask's own tools, in particular in Jupyter. I decided to go with this implementation because the BIFROST workflow already has a progress bar and I don't want to lose it when refactoring to use Sciline more. The BIFROST workflow currently takes ~10min and I think it is important to show that it is making progress. And it makes sense to run such a long process in a terminal outside of a Jupyter notebook for analysis.

The `Reporter` interface proposed here is flexible enough to support other kinds of reporting such as timing providers.

### DIFF
--- a/src/sciline/reporter.py
+++ b/src/sciline/reporter.py
@@ -279,4 +279,8 @@ class _ProviderList:
             active_providers = rf"\[{', '.join(self._provider_names.values())}]"
         else:
             active_providers = ""
+        # Not overflowing the line seems more defensive when this code is used
+        # in a larger context.
+        options.overflow = 'ellipsis'
+        options.no_wrap = True
         return console.render(active_providers, options)
--- a/src/sciline/reporter.py
+++ b/src/sciline/reporter.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+"""Progress reporters for pipelines."""
 
 from __future__ import annotations
 
@@ -14,7 +15,20 @@ from ._utils import provider_name
 
 
 class Reporter(ABC):
-    """Base class for progress reporters of computations."""
+    """Base class for progress reporters of computations.
+
+    A ``Reporter`` instance must not be used for multiple computations concurrently!
+
+    Reporters should be passed to :meth:`sciline.Pipeline.compute`, e.g.:
+
+    .. code-block:: python
+
+         from sciline import Pipeline
+         from sciline.reporter import RichReporter
+         reporter = RichReporter()
+         pipeline = Pipeline(providers)
+         pipeline.compute(Result, reporter=reporter)
+    """
 
     def __init__(self) -> None:
         self._n_steps = 0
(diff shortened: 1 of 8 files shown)


src/sciline/reporter.py
```diff
@@ -0,0 +1,240 @@
...
+        self._provider_id += 1
+        return self._provider_id
+
+
+class RichReporter(Reporter):
+    """Report progress using Rich.
+
+    This class uses the `rich <https://rich.readthedocs.io/en/stable/index.html>`_
+    package to display a progress bar.
+    Rich is not a hard dependency of Sciline and must be installed separately.
```
MridulS, line 103: We should add `rich` to one of the optional deps group (or create a new one). It's easier to keep a track of deps in one single location (pyproject.toml).

src/sciline/reporter.py
```diff
@@ -0,0 +1,256 @@
...
+        .. code-block:: python
+
+            with reporter.run_computation(providers):
+                # Perform computation
+
+        This method exists mainly to provide arguments when entering a reporting
+        context, implementers should mainly override ``__enter__``.
+        """
+        self._n_steps = sum(1 for provider in providers if provider.kind == "function")
+        return self
```
MridulS, line 39: I'm a bit confused here. This is reporting some sort of progress, not exactly the estimated time to completion right? Different providers could take widly different times right?

jl-wynen: Yes, not the estimated time, see https://github.com/scipp/sciline/blob/015e7f275a1153d9ff368889db03251d169fb6e8/src/sciline/reporter.py#L119

The output looks like this (when the computation is completed):
```
Computing ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 7/7 0:00:01
```

src/sciline/reporter.py
```diff
@@ -0,0 +1,256 @@
...
+
+        .. code-block:: python
+
+            with reporter.run_computation(providers):
+                # Perform computation
+
+        This method exists mainly to provide arguments when entering a reporting
+        context, implementers should mainly override ``__enter__``.
+        """
+        self._n_steps = sum(1 for provider in providers if provider.kind == "function")
```
MridulS, line 38: This trips up the case when normal functions are used to create the pipeline. The example at https://scipp.github.io/sciline/#use fails with `AttributeError: 'function' object has no attribute 'kind'` as the task graph isn't created which handles the conversion to `Provider` type.
Or have got the usage wrong here? I tried with:
```python
reporter = sciline.reporter.RichReporter()
with reporter.run_computation(providers=providers):
... 
```

jl-wynen:
I added tests now. I don't know how to test `RichReporter`. But these tests should at least make sure that the reporters are used properly.

src/sciline/reporter.py
jl-wynen: You should not use the context manager yourself. Instead use (adapted from your link)
```python
reporter = sciline.reporter.RichReporter()
providers = [load, clean, process]
params = {Filename: 'data.txt', ScaleFactor: 2.0}
pipeline = sciline.Pipeline(providers, params=params)

pipeline.compute(Result, reporter=reporter)
```

We should never pass `Provider` objects to a pipeline. They are created internally.

PR #199, opened 2025-03-26
Add timing reporter

jl-wynen:
Similarly to the `provider-timings` branch, this wraps providers with timing code. But it can be toggled on and off by users by using the reporter. There is no overhead to this when the reporter is not used.

I used this for benchmarking the BIFROST workflow with different schedulers. Tracking individual times but having a convenient way to see the combined time for each provider was helpful in understanding which providers slowed down the pipeline.


YooSunYoung (approved):

YooSunYoung:
I think this one is worth having a documentation page in the user-guide as well

YooSunYoung:
@jl-wynen The recipe looks really nice! Thanks!

PR #205, opened 2025-04-08
Make Scope variadic

### LINKED ISSUE #203: Consider moving to TypeVarTuple in Python 3.11
Python 3.11 introduced [TypeVarTuple](https://docs.python.org/3.11/library/typing.html#typing.TypeVarTuple) for defining variadic generics. We should be able to use that to make a single `Scope` type instead of two (or maybe more) classes for each number of generic parameters.

I don't know how much this would affect downstream code. It may be that nothing has to change downstream if we also keep `ScopeTwoParams`.

jl-wynen:
Fixes #203

Why wait for 3.11 if this works today?

Amazingly, this needed no changes at all in in `Provider` or `Pipeline`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ requires-python = ">=3.10"
 # Make sure to list one dependency per line.
 dependencies = [
   "cyclebane>=24.06.0",
+  "typing_extensions",  # only for Python < 3.11
 ]
 
 dynamic = ["version"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,7 @@ filterwarnings = [
   "error",
     # From jsonschema
     "ignore:datetime.datetime.utcfromtimestamp:DeprecationWarning",
+    "ignore:ScopeTwoParams is deprecated:DeprecationWarning",
 ]
 
 [tool.ruff]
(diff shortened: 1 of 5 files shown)


jokasimr (commented):
There are some tests using `ScopeTwoParams` that don't seem to be implemented for the new `Scope` that can handle multiple params.

For example this one: 

https://github.com/scipp/sciline/blob/main/tests/utils_test.py#L76-L85

Are those not needed anymore or should we add similar tests using `Scope` with two parameters instead of `ScopeTwoParams`?

jl-wynen:
> There are some tests using `ScopeTwoParams` that don't seem to be implemented for the new `Scope` that can handle multiple params.
> 
> For example this one:
> 
> https://github.com/scipp/sciline/blob/main/tests/utils_test.py#L76-L85
> 
> Are those not needed anymore or should we add similar tests using `Scope` with two parameters instead of `ScopeTwoParams`?

Yes, here: https://github.com/scipp/sciline/blob/a5c1e56387b6d8c2133d40ba5936d3620f4a8022/tests/utils_test.py#L88

Your link points to main. Did you look at the correct file?

jokasimr:
> Did you look at the correct file?

No I did not.

jokasimr (approved):

PR #212, opened 2025-05-12
Allow constraining type vars at construction

jl-wynen:
Motivated by https://github.com/scipp/essreduce/issues/238, this PR allows constraining type vars when building a pipeline. @SimonHeybrock had already started working on this (https://github.com/scipp/sciline/tree/pass-typevar-constraints) but I happened to have the same idea and started work before I saw his comment, so here is my implementation :)

I re-ran the benchmark from https://github.com/scipp/essreduce/issues/238. The proposed change makes the construction times of pipelines independent from the number of constraints in the type var itself. It only depends on the concrete number of constraints passed to `Pipeline.__init__`. (The x axis shows the constraints in the *definition* of the type var. In all cases, it was additionally constrained to 2 types when constructing the pipeline.)
workflow-creation-benchmark-optimised

Note that with this solution, it is still possible to instantiate type vars with bad types and insert such params into the pipeline:
```python
    T = TypeVar('T', int, float, str)

    @dataclass
    class A(Generic[T]):
        v: T

    pipeline = sl.Pipeline(
        [],
        params={A[str]: A[str]("abc")},
        constraints={T: [int, float]},
    )

    assert pipeline.compute(A[str]) == A[str]("abc")
```
Mypy is of no help here because it only sees the constraints in the definition of `T`. This is not ideal but should at worst cause confusion in practice.


jl-wynen:
> Note that with this solution, it is still possible to instantiate type vars with bad types and insert such params into the pipeline:

This also happens with providers:
force-type

I found this because in the generic nexus workflow, the metadata providers require `Filename[SampleRun]`, not `Filename[RunType]`. So this forces `Filename[SampleRun]` into the graph, even when it is excluded by a constraint.

This is not great. But is it worth checking in `insert` and `__setitem__`?

SimonHeybrock:
With this change, can we drop the requirement that typevars are constrained? That would allow for defining "generic" providers, injecting the constraints later, when the pipeline is created?

docs/user-guide/generic-providers.ipynb
```diff
@@ -403,6 +405,58 @@
...
+   "source": [
+    "### Constraining type vars for a specific pipeline\n",
+    "\n",
+    "Sometimes, listing constraints when defining a type variable is too restrictive.\n",
+    "We need to know all possible types that the type variable can take when defining it.\n",
+    "However, there is a workaround.\n",
+    "We can specify constraints when defining a *pipeline* instead of when defining a *type variable*.\n",
+    "\n",
+    "In this example, `T` itself is unconstrained.\n",
+    "But all occurrences of `T` in `pipeline` will be constrained to `int` and `float`."
```
SimonHeybrock, line 421: Should we call this a workaround? I think it makes sense to promote this as a "well" supported mechanism, just like having constraints on the typevar? Maybe it should be mentioned right from the start (where constraints are first mentioned above) that there are two cases?

jl-wynen: Done

SimonHeybrock (approved):

PR #213, opened 2025-05-12
Print path to target when input node is missing

SimonHeybrock:
This avoids printing possibly hundreds of types. I think the old implementation was mainly meant for case 2 below, but case 1 feels more common in practice. We now print ~a little "tree"~ the path from the missing node to the target node, e.g.,:

```
UnsatisfiedRequirement: Missing input node 'Filename'. Affects requested targets (via providers given in parentheses):
1. Filename → (__main__.load) → RawData → (__main__.clean) → CleanedData → (__main__.process) → Result
```

Try it out, e.g., with the "Getting Started" notebook:

Case 1: Remove the `Filename` param and try to compute `Result`
Case 2: Try to compute something not in the graph, such as `int`

### DIFF
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -50,7 +50,9 @@ def _is_multiple_keys(keys: type | Iterable[type] | UnionType) -> bool:
     )
 
 
-def _build_tree_msg(graph, node, depth=0, max_depth=4, indent='    '):
+def _build_tree_msg(
+    graph: Any, node: Any, depth: int = 0, max_depth: int = 4, indent: str = '    '
+) -> str:
     """Build a tree message showing dependencies recursively.
 
     Parameters
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -77,14 +77,12 @@ def _find_paths_to_targets(
     for target in targets:
         try:
             # Find all simple paths from the missing node to the target
-            target_paths = list(all_simple_paths(graph, missing, target))
-            # Sort by length to show the shortest path first
-            paths.extend(sorted(target_paths, key=len))
+            paths.extend(list(all_simple_paths(graph, missing, target)))
         except (nx.NetworkXNoPath, nx.NodeNotFound):  # noqa: PERF203
             # No path found or nodes not in graph
             continue
 
-    return paths
+    return sorted(paths, key=len)  # Sort by length to show the shortest path first
 
 
 def _format_paths_msg(nx_graph: Any, paths: list[list[Any]]) -> str:
--- a/src/sciline/pipeline.py
+++ b/src/sciline/pipeline.py
@@ -50,6 +50,45 @@ def _is_multiple_keys(keys: type | Iterable[type] | UnionType) -> bool:
     )
 
 
+def _build_tree_msg(graph, node, depth=0, max_depth=4, indent='    '):
+    """Build a tree message showing dependencies recursively.
+
+    Parameters
+    ----------
+    graph:
+        The graph to traverse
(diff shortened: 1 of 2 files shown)


jl-wynen:
Why do we need to be guessing here? We have the graph and should be able to exactly identify where a type was needed.

Also, this makes case 2 worse:
```python
from typing import NewType, TypeVar
import sciline as sl

A = NewType('A', int)
B = NewType('B', int)
C = NewType('C', int)
D = NewType('D', int)

def f1(a: A) -> B:
    return B(a + 1)

def f2(b: B, a: A, d: D) -> C:
    return C(b * a)

pl = sl.Pipeline([f1, f2], params={A: 2})
pl.get(D)
```
raises
```
sciline.handler.UnsatisfiedRequirement: Missing input node 'D'. Possibly requested via:
D
    C
```
In this case, it should not walk the graph but just say that `D` is not in the graph.

SimonHeybrock:
> Why do we need to be guessing here? We have the graph and should be able to exactly identify where a type was needed.

Good point! Will fix.

> Also, this makes case 2 worse:
> [...]
> ```
> sciline.handler.UnsatisfiedRequirement: Missing input node 'D'. Possibly requested via:
> D
>     C
> ```
> 
> In this case, it should not walk the graph but just say that `D` is not in the graph.

Yes. Still better than printing 1000 types, but we can make it even better, agreed.

SimonHeybrock:
@jl-wynen Updated according to your suggestion, please have a new look.

src/sciline/pipeline.py
```diff
@@ -50,6 +53,66 @@ def _is_multiple_keys(keys: type | Iterable[type] | UnionType) -> bool:
...
+    -------
+    :
+        The formatted message
+    """
+    msg = f"Missing input node '{key_name(paths[0][0])}'. "
+    msg += "Affects requested targets (via providers given in parentheses):"
+
+    for i, path in enumerate(paths):
+        msg += f"\n{i + 1}. {key_name(path[0])} → "
+        msg += " → ".join(
```
jl-wynen, line 108: I don't know how widely utf-8 is supported in terminals. Especially in embedded terminals in editors and Windows terminals. Can you replace the arrows with `->` or just `>`?

src/sciline/pipeline.py
```diff
@@ -50,6 +53,66 @@ def _is_multiple_keys(keys: type | Iterable[type] | UnionType) -> bool:
...
+        List of paths, where each path is a list of nodes from missing to target
+    """
+
+    paths = []
+    for target in targets:
+        try:
+            # Find all simple paths from the missing node to the target
+            target_paths = list(all_simple_paths(graph, missing, target))
+            # Sort by length to show the shortest path first
+            paths.extend(sorted(target_paths, key=len))
```
jl-wynen, line 82: This can still result in a longer path being shown first. If `targets[0]` produces a path of length 4 and `targets[1]` produces a path of length 2, then we get the length 4 path first in the output. So can you sort the final paths instead of each subset?

jl-wynen (commented):
Nice solution! I am curious to see how useful it will be in practice with large graphs, but it should be an improvement in any case.

src/sciline/pipeline.py
SimonHeybrock: Hmm, we have not avoided utf-8 so far, e.g., for formatting units with "micro"?

jl-wynen: Ok. It won't be fundamentally broken if this can't be rendered. So I am ok with leaving it.

jl-wynen (approved):

PR #215, opened 2025-05-14
Allow constraining typevars to empty set

jl-wynen:
This amends the changes of #212 to allow passing an empty list of constraints when creating a pipeline:
```python
sciline.Pipeline([...], constraints={T: []})
```
This effectively removes all occurrences of `T` from the graph. This can be useful, e.g., with the `GenericNexusWorkflow` when we don't need any monitors. Then we can set `monitor_types=[]` instead of having to specify an arbitrary monitor type that may not even exist at the instrument.


SimonHeybrock (approved):
👍

PR #220, opened 2025-07-31
remove ScopeTwoParams

jokasimr:


YooSunYoung (approved):
I'm assuming we pass the deprecation preiod : D

PR #221, opened 2025-08-06
 drop python 3.10, add python 3.13, copier bump 

MridulS:


nvaytet (approved):
Looks good but Mypy is failing. Should be quite easy to fix though...

PR #222, opened 2025-09-08
Add groupby

nvaytet:
Needs https://github.com/scipp/cyclebane/pull/29

### DIFF
--- a/docs/user-guide/parameter-tables.ipynb
+++ b/docs/user-guide/parameter-tables.ipynb
@@ -229,7 +229,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "graph = pipeline.reduce(func=lambda *result: sum(result), name='merged').get('merged')\n",
+    "def merge(*data):\n",
+    "    return sum(data)\n",
+    "\n",
+    "graph = pipeline.reduce(func=merge, name='merged').get('merged')\n",
     "graph.visualize()"
    ]
   },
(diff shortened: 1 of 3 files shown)


docs/user-guide/parameter-tables.ipynb
```diff
@@ -37,19 +37,19 @@
     "import sciline\n",
     "\n",
     "_fake_filesytem = {\n",
-    "    'file102.txt': [1, 2, float('nan'), 3],\n",
```
SimonHeybrock, line 40: Can you undo the reformat? This is hard to review.

src/sciline/data_graph.py
```diff
@@ -250,6 +270,40 @@ def visualize_data_graph(self, **kwargs: Any) -> graphviz.Digraph:
...
+    """
+
+    def __init__(
+        self, graph: CbGroupbyGraph, graph_maker: Callable[..., DataGraph]
+    ) -> None:
+        self._cbgraph = graph
+        # We forward the constructor so this can be used by both DataGraph and Pipeline
+        self._graph_maker = graph_maker
+
+    def reduce(self, *, func: Callable[..., Any], **kwargs: Any) -> DataGraph:
```
SimonHeybrock, line 288: Since this is user-facing we should repeat all the args (for auto-complete) and docstrings, instead of just using kwargs.

docs/user-guide/parameter-tables.ipynb
SimonHeybrock: `git checkout -p main -- the/file.ipynb` maybe?

PR #229, opened 2026-02-03
MAINT: Update copier and tox -e deps

MridulS:
- **copier bump**
- **tox -e deps**


docs/index.md
```diff
@@ -73,7 +73,7 @@ Your existing Python code can remain generic and can still be used without Scili
 ## At a glance
 
 :::{include} user-guide/installation.md
-:heading-offset: 3
+:heading-offset: 2
```
nvaytet, line 76: This was apparently needed to avoid raising a warning about heading level jumping from 2 to 4.

nvaytet:
@MridulS Can you take a quick look. I think I fixed most things.

I also ran another update as the copier template had changed since you last opened the PR.

I also changed the CI slightly so that the `package-test` is only run once (note that I copied the setup over from Plopp).

nvaytet (approved):

PR #231, opened 2026-05-28
Discard data early in the naive scheduler

jl-wynen:
This reproduces the mechanism from `scipp.transform_coords` to discard intermediate results once they are no longer required. This is not the optimal solution because `TopologicalSorter` does not optimise for this. But in most cases, this will still reduce peak memory usage..

### DIFF
--- a/src/sciline/scheduler.py
+++ b/src/sciline/scheduler.py
@@ -1,11 +1,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
 import inspect
+from collections import Counter
 from collections.abc import Callable, Hashable
 from typing import Any, Protocol, runtime_checkable
 
 from sciline.typing import Graph
 
+from ._provider import Provider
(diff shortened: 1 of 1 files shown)


SimonHeybrock (commented):
Looks ok. Given that `TopologicalSorter` is apparently implemented in a BFS-like manner this is likely far from optimal but better than nothing.

SimonHeybrock (approved):

jl-wynen:
> Looks ok. Given that `TopologicalSorter` is apparently implemented in a BFS-like manner this is likely far from optimal but better than nothing.

BFS might be best here because that means that caches values are deleted as soon as possible.

SimonHeybrock:
> > Looks ok. Given that `TopologicalSorter` is apparently implemented in a BFS-like manner this is likely far from optimal but better than nothing.
> 
> BFS might be best here because that means that caches values are deleted as soon as possible.

But doesn't it mean we are potentially keeping more nodes in parallel, or more raw-data nodes that might be larger than what we'd get if we followed the chain as far down as possible?

Issue #233, opened 2026-07-30
Scope does not work with Python 3.12 generics


doronbehar:
Test failures we experience on NixOS:

```
============================= test session starts ==============================
platform linux -- Python 3.14.6, pytest-9.1.1, pluggy-1.6.0 -- /nix/store/rgnappqqc5vbq60gza5fflyk84sylwl6-python3-3.14.6/bin/python3.14
cachedir: .pytest_cache
Using --randomly-seed=17394782
rootdir: /build/source
configfile: pyproject.toml
testpaths: tests
plugins: randomly-4.0.1
collected 231 items / 1 error

==================================== ERRORS ====================================
_______________ ERROR collecting tests/complex_workflow_test.py ________________
tests/complex_workflow_test.py:32: in <module>
    class Masked(sl.Scope[Run, npt.NDArray[np.float64]], npt.NDArray[np.float64]): ...
E   TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
=========================== short test summary info ============================
ERROR tests/complex_workflow_test.py - TypeError: metaclass conflict: the metaclass of a derived class must be a (...
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.85s ===============================
```

Even when we added the `--ignore-glob=tests/complex_workflow_test.py` Pytest flag, we got the error:

```
=================================== FAILURES ===================================
______ test_subclasses_of_generic_array_provider_defined_with_Scope_work _______

    def test_subclasses_of_generic_array_provider_defined_with_Scope_work() -> None:
        # int is unused, but a single constraint is not allowed by Python
        Param = TypeVar('Param', str, int)
    
>       class ArrayT(sl.Scope[Param, npt.NDArray[np.int64]], npt.NDArray[np.int64]): ...
E       TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases

tests/pipeline_test.py:256: TypeError
=========================== short test summary info ============================
SKIPPED [1] tests/pipeline_test.py:1269: Test requires dask.distributed
SKIPPED [1] tests/pipeline_test.py:1286: Test requires dask.distributed
SKIPPED [1] tests/pipeline_test.py:1259: Test requires dask.distributed
[...]

PR #234, opened 2026-08-18
Use np.ndarray in generic types

jl-wynen:
The previous npt.NDArray[...] clashes with `typing.Generic` coming from `sciline.Scope`. Functionally, there is no reason to inherit from or parametrise by npt.NDArray.

This fixes the immediate error in #233. This will not currently be tested with a version that triggers the failure because I do not want to add a new Python 3.12 test since we will soon drop support for 3.11 anyway.


MridulS (approved):

SimonHeybrock:
Thanks, should #233 be closed?

jl-wynen:
> Thanks, should #233 be closed?

Depends on whether you want to figure out how to handle py3.12 generics. Or if we should just leave everything as is and accept that users can run into this error.

Issue #235, opened 2026-08-20
Support for Python 3.12 generics


jl-wynen:
Python 3.12 introduced generic functions and classes. In principle, they can replace our custom `Scope`. Doing so would fix #233. But this does not currently work well with our specialisation mechanism. For example:

```python
import sciline as sl

type A = int
type B = int
class Raw[Run](float): ...
class Processed[Run](float): ...

def foo[Run](x: Raw[Run]) -> Processed[Run]:
    return Processed[Run](x * 2)

pl = sl.Pipeline([foo], params={Raw[A]: Raw[A](1.2)})
```
The above fails with
```
ValueError: Type variable Run has no constraints. Either constrain the type variable in its definition or via the 'constraints' argument of Pipeline.
```
Adding constraints via
```python
pl = sl.Pipeline(
    [foo],
    params={Raw[A]: Raw[A](1.2)},
    constraints={Raw: (A, B)},
)
```
changes the error to
```
AttributeError: type object 'Raw' has no attribute '__constraints__'
```

Unfortunately, it is not as simple as fixing how constraints are extracted. The code above has 3 distinct TypeVars called `Run`:
```python
for param in [
    Raw.__type_params__[0],
    Processed.__type_params__[0],
    foo.__annotations__["x"].__parameters__[0]
]:
    print(param, type(param), id(param), param == Raw.__type_params__[0]))
```
```
Run <class 'typing.TypeVar'> 140284653425744 True
Run <class 'typing.TypeVar'> 140284633285488 False
Run <class 'typing.TypeVar'> 140284632928704 False
```
As the example shows, the 3 TypeVars do not compare equal with each other. So our mechanism of matching TypeVars across providers does not work.

**Alternative to class def**

There is an alternative way to define generic type aliases:
```python
type Raw[Run] = float
```
But it has the same problem; `Run` is a unique TypeVar, local to `Raw`.

### Current workaround

It is already possible to make this work by explicitly listing constraints in *every* generic:
```python
import sciline as sl

type A = int
type B = int

class Raw[Run: (A, B)](float): ...
class Processed[Run: (A, B)](float): ...

def foo[Run: (A, B)](x: Raw[Run]) -> Processed[Run]:
    return Processed[Run](x * 2)

pl = sl.Pipeline(
    [foo],
    params={Raw[A]: Raw[A](1.2)},
)
```
But this does not scale well. And it does not work with constraints defined in technique packages.

### Possible solution

[...]

PR #236, opened 2026-08-21
Prototype: demand-driven instantiation of generic providers

SimonHeybrock:
Exploration for #235 — a draft for discussion, not a final proposal or decision.

## Idea

The only place where sciline relies on `TypeVar` identity *across* definitions is the constraints lookup. Within a single provider, PEP 695's scoped type variables are self-consistent: all annotations of `def foo[Run](x: Raw[Run]) -> Processed[Run]` share one `Run` object, which is all the existing per-provider binding needs.

So instead of requiring constraints to enumerate instantiations eagerly at insertion, this prototype keeps generic providers with unconstrained type variables as templates and instantiates them on demand, by unifying their type patterns with the concrete keys that appear in the pipeline:

- backward from requested targets (`get`, `compute`, `__getitem__`): `Processed[A]` unifies with the return pattern `Processed[Run]`, binding `Run=A`, recursing into the resulting dependencies;
- forward from present concrete keys where the concrete graph is needed before targets are known (`map`, `output_keys`, no-arg `visualize`).

With this, the example from the issue works verbatim, with no constraints declared anywhere:

```python
import sciline as sl

type A = int
type B = int
class Raw[Run](float): ...
class Processed[Run](float): ...

def foo[Run](x: Raw[Run]) -> Processed[Run]:
    return Processed[Run](x * 2)

pl = sl.Pipeline([foo], params={Raw[A]: Raw[A](1.2)})
pl.compute(Processed[A])  # Processed[A](2.4)
```

Providers whose type variables *are* constrained (the `class Raw[Run: (A, B)]` workaround, old-style constrained `TypeVar`s, and the `constraints=` argument) are expanded eagerly exactly as before.

## What works

See `tests/pep695_test.py`: provider chains; source providers with no generic inputs (instantiated from targets alone); shadowing by concrete providers and by params; multi-TypeVar providers with transitive instantiation; `type Raw[Run] = float` aliases; `map` over a purely generic pipeline (forward chaining from the mapped keys); values for generic keys (`params={Raw: Raw(1.2)}`), stored as templates and applied on demand to all requested specializations; old-style unconstrained `TypeVar`s now work the same way, making `constraints=` optional in general. Unsatisfied-requirement error messages reflect the instantiated graph.

## Semantic changes and open questions

- Unconstrained type variables no longer raise at insertion; failure moves to graph build as an unsatisfied requirement (one existing test changed accordingly). Less fail-fast, though the build-time error is clear.
- When several generic providers match a requested key, the last-inserted one wins, mirroring concrete-provider replacement. There is no "most specific wins" for overlapping patterns such as `Processed[R]` vs `Processed[list[R]]`.
- Should `constraints=` stay (e.g. as validation/restriction of what inference may instantiate), or be deprecated?
- Forward chaining uses union semantics per argument, which can over-instantiate. This bit once already: unrestricted expansion at `map()` time instantiated a provider for an unrelated concrete key, creating an extra sink that broke `reduce()`'s unique-sink requirement. Expansion at `map()` is now restricted to bindings consuming a mapped key or a key derived from one; other cases remain demand-driven.
- Precedence among templates is: concrete providers/params, then generic values, then generic providers (latest wins within each). This differs from the last-write-wins semantics of concrete keys.
- `_repr_html_` still shows the unexpanded graph; forward chaining is a naive fixed-point iteration, O(templates × nodes) per round.
- Side observation: `_from_cyclebane` now propagates `_constraints` through `map`/`reduce`/`__getitem__`; previously they were dropped on those paths (only `copy()` preserved them), which looks like a latent bug.

### DIFF
--- a/src/sciline/_unification.py
+++ b/src/sciline/_unification.py
@@ -64,35 +64,37 @@ def match_return(template: Provider, key: Key) -> Provider | None:
 
 def forward_bindings(
     template: Provider, known_keys: Iterable[Key]
-) -> Generator[dict[TypeVar, Key], None, None]:
+) -> Generator[tuple[dict[TypeVar, Key], frozenset[Key]], None, None]:
     """Enumerate complete bindings of a template's TypeVars from known keys.
 
     Each generic argument of the template is unified with each known key;
     consistent combinations of the resulting bindings that bind all type
-    variables of the template are yielded. Arguments that match no known key
-    are left unmatched, i.e., a binding is complete as long as the *other*
-    arguments determine all type variables.
+    variables of the template are yielded, together with the set of known keys
+    that produced them. Arguments that match no known key are left unmatched,
+    i.e., a binding is complete as long as the *other* arguments determine all
+    type variables.
     """
     typevars = find_all_typevars(template.deduce_key())
     patterns = [p for p in template.arg_spec.keys() if find_all_typevars(p)]
     options = []
     for pattern in patterns:
-        matches: list[dict[TypeVar, Key]] = [{}]
+        matches: list[tuple[dict[TypeVar, Key], Key | None]] = [({}, None)]
         for key in known_keys:
             bound: dict[TypeVar, Key] = {}
             if unify(pattern, key, bound):
-                matches.append(bound)
+                matches.append((bound, key))
         options.append(matches)
     seen = set()
     for combo in itertools.product(*options):
         merged: dict[TypeVar, Key] = {}
-        if not all(_merge(merged, bound) for bound in combo):
+        if not all(_merge(merged, bound) for bound, _ in combo):
             continue
         if set(merged) != typevars:
             continue
-        if (fingerprint := frozenset(merged.items())) not in seen:
+        used = frozenset(key for _, key in combo if key is not None)
+        if (fingerprint := (frozenset(merged.items()), used)) not in seen:
             seen.add(fingerprint)
-            yield merged
+            yield merged, used
 
 
 def _merge(target: dict[TypeVar, Key], bound: dict[TypeVar, Key]) -> bool:
--- a/src/sciline/_unification.py
+++ b/src/sciline/_unification.py
@@ -1,0 +1,103 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2025 Scipp contributors (https://github.com/scipp)
+"""Unification of generic type patterns with concrete keys.
+
+Generic providers whose type variables lack constraints are not expanded
+eagerly. They are kept as templates and instantiated on demand by unifying
+their argument and return-type patterns with the concrete keys that appear in
+the pipeline (parameters, requested targets, mapped keys).
+"""
+
+from __future__ import annotations
+
+import itertools
+from collections.abc import Generator, Iterable
+from typing import TYPE_CHECKING, TypeVar, get_args, get_origin
+
+if TYPE_CHECKING:
+    from ._provider import Provider
+    from .typing import Key
+
+
+def find_all_typevars(t: type | TypeVar) -> set[TypeVar]:
+    """Returns the set of all TypeVars in a type expression."""
+    if isinstance(t, TypeVar):
+        return {t}
+    if params := getattr(t, '__parameters__', ()):
+        return set(params)
+    return set(itertools.chain(*map(find_all_typevars, get_args(t))))
+
+
+def unify(pattern: Key | TypeVar, concrete: Key, bound: dict[TypeVar, Key]) -> bool:
+    """Match ``concrete`` against ``pattern``, extending ``bound`` in place.
+
+    Returns True on success. ``bound`` may contain partial bindings on failure
+    and must be discarded by the caller in that case.
+    """
+    if isinstance(pattern, TypeVar):
+        if pattern.__constraints__ and concrete not in pattern.__constraints__:
+            return False
+        if pattern in bound:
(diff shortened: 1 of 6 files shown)


src/sciline/_unification.py
jl-wynen, line 1: This breaks this example:
```python
pl = sl.Pipeline([foo], params={Raw: Raw(1.2)})
print(pl.compute(Processed[A]))
```
On main, we can insert generic parameters that apply to all specialisations.

This means that 
> __setitem__ with a generic key still requires constraints, since there is nothing to enumerate from.

is anyway broken.

jl-wynen, line 1: > So instead of requiring constraints to enumerate instantiations eagerly at insertion, this prototype keeps generic providers with unconstrained type variables as templates and instantiates them on demand, 

We need to be really careful here: Data graphs were designed to be concrete graphs without generics. So changing this may break assumptions. And it may break the fundamental design of `DataGraph`.

SimonHeybrock: Good catch. Support for this is now in 2fc0461: a value set for a generic key is stored as a template and matched on demand against requested specializations, symmetric to generic providers — the example above works without constraints. One semantic difference worth noting: precedence is concrete providers/params, then generic values, then generic providers, rather than the last-write-wins order concrete keys have.

SimonHeybrock: Agreed, this is the main risk. The prototype tries to preserve the invariant in a narrower form: the cyclebane graph itself stays concrete at all times — templates live outside it in a side table, and expansion happens before cyclebane sees any operation (`map`) or on a copy at build time (`get`, `__getitem__`). So cyclebane never encounters a generic node. What the design loses is a different guarantee: the graph is only *complete* after expansion, so every current and future call site that reads the graph must remember the expansion hook, and a forgotten hook fails silently by seeing fewer nodes rather than erroring.

Your concern was borne out once already in a different way: unrestricted forward expansion at `map()` over-instantiated a provider for an unrelated concrete key, and the extra sink broke `reduce()`'s unique-sink assumption (fixed in 2fc0461 by restricting expansion to bindings that consume mapped keys). I'll post a comment on the thread with a fuller list of what coexistence of the eager and demand-driven paths costs.

SimonHeybrock:
Some notes on what maintaining the eager (constraints-based) and demand-driven paths in parallel would cost. It goes deeper than `constraints=` becoming optional:

**Divergent semantics, not just duplicate code.** A single pipeline can contain both kinds of provider, and the two paths answer the same questions differently:

- *Replacement/shadowing.* On the eager path insertion order decides: a generic provider expanded at insert claims nodes for all constraint values, and whoever inserts last for a key wins — a generic provider can overwrite a concrete one. On the template path, templates never overwrite an existing satisfied node, so a concrete provider always shadows a generic one regardless of order. `Pipeline([concrete, generic])` and `Pipeline([generic, concrete])` behave identically in one world and differently in the other.
- *Node existence.* Eager generics materialize the full cross-product at insert; templates materialize nothing until demanded. This leaks into every API that inspects the graph: `_repr_html_`, `visualize_data_graph`, `output_keys`, `get_mapped_node_names`, direct iteration over `underlying_graph`. In a mixed pipeline some generics are "in" the graph and others invisible.
- *Failure timing.* Eager fails at construction (bad constraint values, unsatisfiable enumeration); inference fails at graph build as an unsatisfied requirement. Same conceptual mistake, two error moments.
- *Mixed providers.* A provider with one constrained and one unconstrained type variable lands on the template path, where the declared constraints are enforced by `unify()` rejecting bindings rather than by enumeration — a third, hybrid behavior.

**An ongoing invariant tax.** The demand hooks (`__getitem__`, `map`, `to_task_graph`, `output_keys`, the error-message path) exist only because of templates. Every future API that consumes concrete keys has to remember to expand first, and a forgotten hook does not error — it silently sees an unexpanded graph. This is the flip side of the concern about `DataGraph`'s concrete-graph design raised in the review: the cyclebane graph stays concrete, but "concrete" is replaced by the weaker invariant "complete only after expansion".

**Doubled surface.** Tests for generics-related changes need eager, inference, and mixed variants; docs must teach both models and the differences above — for users this is probably the largest cost.

**Possible consolidation.** Much of the implementation duplication could be collapsed by making eager a special case of inference: treat declared constraints as seed demand, i.e. register the provider as a template and immediately expand it for the constraint combinations at insert. That would leave one mechanism and one shadowing rule, while constrained code keeps its visible behavior (nodes exist at insert, errors fire early). What consolidation cannot remove are the user-visible model differences — node-existence and error timing — which persist as long as both styles are supported. Realistically that is a while, since ess* currently uses constrained `Scope`-style type variables throughout.

SimonHeybrock:
#237 explores the other end of the coexistence question discussed above: removing the eager path entirely, so demand-driven instantiation is the only mechanism. It is based on this branch, so its diff shows exactly what the consolidation removes (net −118 lines in the engine). The description there includes measured compatibility data from running the existing test suite against the single-model engine: 242 of 248 tests pass unchanged, six tested the removed `constraints=` argument, and one semantic change remains (specialized providers shadow generic ones regardless of insertion order). Both PRs are proposals for discussion, not decisions.

SimonHeybrock:
Status update: #237 has been retargeted to `main` and now carries the consolidated end-state proposal (single mechanism, backward-only, deferred mapped-labeling via scipp/cyclebane#32); #238 was merged into it. This PR remains open as the coexistence alternative — i.e., the fallback in case review favors keeping the eager path alongside demand-driven instantiation.

PR #238, opened 2026-08-21
Prototype: demand-driven generics without map-time forward instantiation

SimonHeybrock:
Q1 spike from the design document (`docs/developer/architecture-and-design/demand-driven-generics.md`): remove the map-time forward-instantiation hook by relying on cyclebane deriving mapped-node labeling at task-graph build time. **Depends on scipp/cyclebane#32** — CI will be red until that lands and is released. Based on the #237 branch; draft for discussion.

## What this shows

With cyclebane keeping plain node names and deriving which nodes carry which indices at `to_networkx()` time, `map()` commutes with provider insertion: generic providers instantiated on demand *after* mapping receive their indices from the derivation. Consequences here:

- The `map()` hook and the entire seed-restriction machinery are gone; forward chaining survives only behind `output_keys()`, whose fate is Q3 (rule-graph inspection).
- Mapped roots count as satisfied keys, so backward instantiation never tries to provide them.
- `reduce()` with an explicit `key` treats the key as a demand and instantiates providers for it. Without a key, the reduced sink is resolved among nodes actually present, so reducing a rule-derived sink requires an explicit `key` — two tests updated accordingly. This is the user-visible rule to review.
- `get_mapped_node_names` reads derived indices via cyclebane's new `named_indices` accessor instead of scanning for `MappedNode` labels, and loses the multiple-candidates ambiguity handling (plain names are unique).

All 245 tests pass (against the cyclebane branch); notably, before this change, the whole suite already passed against the cyclebane branch *without any sciline modification*, i.e. the cyclebane change is transparent to current sciline.

## Discussion points

- The reduce-`key` rule above: acceptable, or should sciline defer sink resolution further?
- The shadowing seams (reduce results and grafted branches reusing a mapped node's name) keep a scoped `MappedNode` alias inside cyclebane — see the discussion in scipp/cyclebane#32.


SimonHeybrock:
Follow-up per discussion (4cbf9fd): forward chaining is now fully deleted (`_instantiate_forward`, `forward_bindings`) — the engine is backward-only. `output_keys()` returns concrete sinks plus return-type patterns of generic providers not consumed by other generic providers; no-arg `visualize()` shows the concrete part of the graph. With scipp/cyclebane#32 forbidding the shadowing idioms, `get_mapped_node_names` loses the multiple-candidates disambiguation (`index_names=` is pure validation now) and the affected tests use distinct names for a mapped node and its reduction. Decisions recorded in the design document.

SimonHeybrock:
Merged into #237 (fast-forward, commits preserved), which is now the consolidated end-state proposal against main. See the design document there for the decision log.

Issue #241, opened 2026-08-31
Feeding values into a graph repeatedly requires a rebuild and retains the values


SimonHeybrock:
## Problem

Repeatedly computing the same graph with changing input values — streaming, fitting, scanning — currently has to assign the values: `pipeline[key] = value`. Each assignment goes through `DataGraph.__setitem__` → `cyclebane.Graph.__setitem__`, i.e. a full graph rebuild, although the graph *structure* does not depend on the values. The assigned value also stays reachable from the pipeline until the next assignment overwrites it, which for large arrays means an extra live buffer per key.

`ess.reduce.streaming.StreamProcessor` does this once per dynamic key per chunk, once per accumulator per finalize, and twice per context key plus once per derived node per context update, all at roughly 1 Hz per running job. Measured consequences (scipp/ess#732):

- Memory: on cyclebane before scipp/cyclebane#33, a minimal cumulative-histogram processor with 16 MB chunks grew 138 → 747 MB over 20 cycles, because the discarded graphs hold the arrays in node attribute dicts and are reclaimable only by the cyclic collector. This OOM-killed a production service (scipp/esslivedata#1264).
- Steady state: one extra chunk plus one extra accumulator value are held at all times (scipp/esslivedata#1259).
- CPU: across the 41 real esslivedata workflows, the rebuilds cost a median of 0.61 ms and up to 3.54 ms per accumulate+finalize cycle, plus up to about 9 ms per context update for workflows whose context comes from f144 device logs.

scipp/ess#732 works around this inside essreduce: a mapping the workflow reads its inputs from, fed by a provider inserted per key, so only the first value for a key is assigned and later ones are dict writes. It works, but it patches `__annotations__` at runtime to tell sciline which key a provider provides, which is a sign the mechanism belongs here rather than downstream.

## What would help

Two options, of quite different value.

**A `compute(targets, params={...})`-style API.** Removes the rebuild and gives values a call-scoped lifetime. Roughly equivalent to what scipp/ess#732 already achieves, so the gain is that the mechanism is upstream and the annotation patching disappears.

**A reusable task graph — built once, values supplied per call.** This is the one worth doing. Under the naive scheduler `Pipeline.get()` accounts for roughly 80% of a `compute()` call (0.52 ms out of 0.64 ms for a 30-node chain), which neither scipp/ess#732 nor a bare `params=` touches. Doing it outside sciline would mean constructing a `Provider.parameter` and writing into `TaskGraph._graph`, two private layers deep.

[...]

PR #245, opened 2026-09-11
ADR 0003: Stages and aggregations outside the graph

SimonHeybrock:
Proposes replacing `Pipeline.map`/`reduce` with two objects built from an ordinary flat pipeline, outside the graph: a `Stage`, the part of a pipeline from named input keys to named output keys with the static part computed once and held, and an `Aggregation`, two stages with an accumulator per accumulation key between them. This PR is the additive half: the ADR (status proposed), the design document with a rollout plan, the two modules, their tests, and a user-guide page. `map`, `reduce`, and cyclebane are untouched; nothing breaks.

## Why

Map/reduce inside the graph has been implemented twice and is the source of every breaking change and open question in the PEP 695 generics prototypes (#236, #237). Three consumers need an operation it cannot express: run one part of a graph repeatedly with some values supplied per call, combine the results outside, run the rest once. `ess.reduce.streaming.StreamProcessor` builds that partition by hand (#241 records the requirements), the essapps reduction-service architecture needs it in three places, and every ESS reduction package folds runs at one or more keys and grafts the result back onto the pipeline.

The ADR states the decision, the alternatives tried and rejected during prototyping, and the consequences. The design document has the semantics, the survey of every map/reduce site in the ESS packages and esslivedata, the evidence, and a PR-level rollout plan across sciline, ess.reduce, the reduction packages, esslivedata, and essapps.

## What is in the PR

- `docs/developer/adr/0003-replace-map-reduce-with-stages-outside-the-graph.md`: the decision. Read this first.
- `docs/developer/architecture-and-design/map-reduce-outside-the-graph.md`: design, survey, evidence, migration, rollout plan, settled and open questions.
- `sciline.Stage`, `sciline.warm`: the graph partition.
- `sciline.Accumulator` (protocol), `sciline.Aggregation`, `sciline.compute_members`: the table-fold shape and the replacement for `compute_mapped`.
- `sciline.Buffered` and `sciline.Reduced`: accumulator factories for an n-ary function applied to all pushed values, and for a running result of an associative binary function. `Reduced` is the memory-cheap choice for sums of large arrays; it never updates in place, so it cannot corrupt values the caller still holds.
- `docs/user-guide/stages-and-aggregations.ipynb`: a guide next to the parameter-tables page. It does not claim to replace anything while the ADR is proposed, but reviewers may find it the quickest way to see the semantics.
- `docs/developer/architecture-and-design/loki_validation.py`: the esssans multi-run reduction over stages against the map/reduce reference. Results identical, provider call counts identical.
- `scheduler_or_default` in `sciline.scheduler`, shared by `TaskGraph` and `Stage`.

Tracking issues: scipp/sciline#246 for the sciline side, scipp/ess#746 for ess.reduce, the reduction packages, and esslivedata.

## Not in the PR

`provide(key, callable)`, a `reporter` argument on `Stage`, and visualization over stages are listed as a follow-up in the rollout plan. The breaking release that removes `map`/`reduce` waits until the ESS packages and esslivedata have migrated; the plan gives the order.

## Test plan

- New tests in `tests/stage_test.py` and `tests/aggregation_test.py`; the full suite passes.
- mypy strict is clean on the new modules.
- `loki_validation.py` run locally against the reference: identical results and call counts.
- The user-guide notebook executes cleanly under nbconvert.


docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,418 @@
...
+- `Accumulator`, in sciline: the structural protocol for what sits between stages, `push` a value and read the combined `value`. `Buffered(func)` is a factory for an accumulator that holds all pushed values and applies an n-ary function on `value`; `Reduced(func)` is a factory for one that holds only a running result of an associative binary function. The accumulators of `ess.reduce.streaming` satisfy the protocol. `Forwarder`, which holds the latest value pushed, is the connector for a context between stages and stays in ess.reduce.
+- `Aggregation`, in sciline: the table-fold shape as two stages of one pipeline, contribute and finalize, with an accumulator per accumulation key between them and the three entry points exposed. It holds nothing but its stages; parameters are set on the pipeline, and whoever loops over members owns the contributions.
```
jl-wynen: I don't understand this paragraph. What is a 'table-fold shape'? What are the three entry points?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+- `Stage`, in sciline: the part of a pipeline from a set of input keys to a set of output keys.
+  The stage computes once everything that does not depend on the inputs, and the caller supplies the inputs per call.
+  This is what `StreamProcessor` builds by hand today, what scipp/sciline#241 asks for, and what the essapps warm workflow (D8) and split workflows (phase 3) are.
+- `Accumulator`, in sciline: the structural protocol for the object between two stages.
+  A caller pushes a value with `push` and reads the combined `value`.
+  `Buffered(func)` is a factory for an accumulator that holds all pushed values and applies an n-ary function on `value`.
+  `Reduced(func)` is a factory for an accumulator that holds only a running result of an associative binary function.
+  The accumulators of `ess.reduce.streaming` satisfy the protocol.
+  `Forwarder`, which holds the value of the latest push, is the connector for a context between stages and stays in ess.reduce.
+- `Aggregation`, in sciline: the table-fold shape as two stages of one pipeline, contribute and finalize, with one accumulator per accumulation key between them and with the three entry points exposed.
```
jl-wynen: What is a 'table-fold shape'? And which three entry points?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+  It holds nothing but its stages.
+  The parameters are set on the pipeline, and whoever loops over the members owns the contributions.
+- `StreamProcessor`, in ess.reduce, as a loop over stages and connectors that carries only its policy.
+
+There is no drop-in replacement for the pipeline that `with_sample_runs` and the related `with_*` helpers return today.
+Each package returns its own object, about thirty lines over `Aggregation`, which holds the pipeline, the aggregations, the contributions, and the "set runs, set parameters, compute" experience.
```
jl-wynen: What does 'about thirty lines over Aggregation' mean?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+A prototype on sciline `main` passes 27 tests that cover the use-case shapes of the ESS packages.
+It reproduces the LoKI multi-run reduction identically against the existing `with_sample_runs`, with the same provider call counts.
+
+## Problem
+
+### In sciline
+
+`map` relabels every reachable node at call time, before the targets are known.
+The demand-driven generics work paid for that twice.
+First with forward chaining and its seeding, which the Q1 spike removed with the deferral of the labeling into cyclebane (scipp/cyclebane#32, unreleased, which is why the CI of the branch is red).
```
jl-wynen: I still have no clue what a 'Q1 spike' is supposed to be.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+
+There is no drop-in replacement for the pipeline that `with_sample_runs` and the related `with_*` helpers return today.
+Each package returns its own object, about thirty lines over `Aggregation`, which holds the pipeline, the aggregations, the contributions, and the "set runs, set parameters, compute" experience.
+
+A prototype on sciline `main` passes 27 tests that cover the use-case shapes of the ESS packages.
+It reproduces the LoKI multi-run reduction identically against the existing `with_sample_runs`, with the same provider call counts.
+
+## Problem
+
+### In sciline
```
jl-wynen: For future reference, I think it needs to be clear that this is not a problem with Sciline but rather a problem with the proposed changes. This might currently read as through the released version of Sciline has these problems (which are implementation details)

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+Then with what that deferral cost: the `reduce(key=)` sink guard, `_consumed_by_template`, the mapped-root case in `_satisfied`, the break of the node-name uniqueness that forbids the `pipeline[C] = pipeline[C].map(...).reduce(...)` idiom that every ESS package uses, and the rework of `get_mapped_node_names`.
+Those are the breaking changes of scipp/sciline#237, and every one of them is attributable to map/reduce.
+Users cannot write a mapped key.
+They must use `get_mapped_node_names` and `compute_mapped`, which need pandas and the node-name classes of cyclebane.
+Sciline never exposed `groupby`, and nothing downstream uses `index_names` or `indices`.
+
+### In ess.reduce
+
+`ess.reduce.streaming` is 1080 lines, about half of it `StreamProcessor`.
+Its core is a graph partition, computed once: the ancestors of the targets, the descendants of the dynamic keys, and the frontier between them.
```
jl-wynen:
```suggestion
Its core is a graph partition, computed once: the ancestors of the targets, the descendants of the dynamic keys, and the boundary between them.
```

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+It therefore assigns `None` to keys to prune branches, grafts subgraphs through `__setitem__`, and, after scipp/ess#732, feeds values through providers whose `__annotations__` are patched at run time.
+scipp/sciline#241 records the three requirements that the mechanism must meet: no rebuild of the graph per value, an explicit lifetime for the supplied values, and removal of what the ancestors of a supplied key would otherwise compute.
+
+### In essapps
+
+Three parts of the architecture are the same partition:
+
+- D8, the warm workflow: cache the nodes just upstream of the cheap parameters, and rerun what lies downstream.
+- D15, the declared additive combine: the graph up to the accumulation keys is contribute, the graph from the keys to the targets is finalize, and combine is outside both.
+- Phase 3, the split workflow: the same partition, but the value at the boundary crosses a spec boundary as a stored stage output.
```
jl-wynen: What are 'D8', 'D15', 'Phase 3'?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+
+The architecture notes that `StreamProcessor` "is the additive structure made explicit" and that the wrapper "gets the three from the accumulation keys".
+It does not yet have a name for the thing that makes the partition.
+
+## What map/reduce is used for
+
+Every use in `/workspace/ess` and esslivedata, by shape:
```
jl-wynen: Reference to local filesystem

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+
+Tests: 238 today; 16 use map/reduce and go, the map-related tests of the PEP 695 branch are trimmed, and the rest port.
+Docs: a guide on stages and aggregations replaces the parameter-tables guide, and the generic-providers guide loses `constraints=`.
+
+### 2. `Stage`
+
+```python
+stage = Stage(pipeline, outputs=(Numerator, Denominator), inputs=(Filename,))
+stage.frontier          # the static keys the dynamic part reads, computed once
+values = stage({Filename: 'run1.nxs'})   # -> {Numerator: ..., Denominator: ...}
```
jl-wynen: Why is this a function call? `Stage` is more than just a function, so I think this should be a named method. Like `compute` for parity with `Pipeline`.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+`Aggregation` needs nothing beyond `Stage` and the `Accumulator` protocol, but it belongs in sciline too, because sciline documents map/reduce for all users and `Aggregation` is its replacement; see Migration.
+
+### 3. Composition: stages and connectors
+
+Every use above is a set of stages of one flat pipeline with something between them, and that something is where the state and the policy live.
+Between two stages sits a connector, an object with `push` and `value`:
+
+- an accumulator, where the values of the members are combined.
+  `Accumulator` is a structural protocol in sciline, with `push(value)` and a `value` property, because `Aggregation` consumes it.
+  `Buffered(func)`, in sciline, is a factory for an accumulator that holds all pushed values and applies an n-ary function on `value`.
```
jl-wynen: This sounds like the function called in `value` can be arbitrarily expensive. So `value` should not be a property.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+  `Reduced(func)`, in sciline, is a factory for an accumulator that holds only a running result and applies an associative binary function to that result and to each pushed value.
+  The function must not modify its arguments, since the pushed values belong to the caller.
+  The accumulators of `ess.reduce.streaming` satisfy the protocol as they are, once `maybe_hist` moves out of their base class.
+  `clear` stays a convention of ess.reduce for accumulators that are reused between finalizes.
+- a forwarder, where a value is held until it is replaced: the context of `StreamProcessor`, the frontier of a warm workflow, or a stage output that crosses a process boundary in essapps phase 3.
+  `Forwarder` stays in ess.reduce, because nothing in sciline consumes one.
+- a dict by member label, where the caller that loops over the members holds the contributions.
+
+The accumulator decides whether a combine buffers or runs incrementally.
```
jl-wynen: This sentence makes no sense. Should it be 'to combine'? Or is 'combine' supposed to be a noun here?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+The aggregation does not decide this.
+In both cases `combine` takes combined values as well as contributions, and thus the combine can proceed in groups or as a chain.
```
jl-wynen: Again 'combine' as a noun? Also, what does proceeding in groups mean here?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+That asks two things of an accumulator: `push` must accept what `value` returns, and the result must not depend on the groups of the pushes.
```
jl-wynen: What groups of pushes?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+`Buffered` has both if its function is associative, `Reduced` requires it, and an accumulator of ess.reduce has them if `push` accepts what `value` returns, which the histogramming ones must keep once `maybe_hist` moves out of the base class.
+For concat-like combines the two cost about the same, twice the total, and `Buffered` is the correct choice.
```
jl-wynen: I don't understand. What 'two' and what 'total'? And how can total be less than its constituents?

---

I think this paragraph is too long and uses too many relative expressions that make it hard to understand what each sentence refers to. Do we even need this here? It does not seem important to the design itself.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+For a sum over large dense arrays a running total holds one array instead of one array per member, and an incremental accumulator such as `Reduced` is the correct choice.
+An accumulator in the table case and in the stream case is the same object.
+Whether the values arrive from one input over time or from many members is the picture of the driver.
+Lifetime (`clear`), identity (contributions by label), and the dependence on the order belong to the driver.
+
+`Stage` itself holds nothing but its frontier, and even that frontier is a forwarder from a stage without inputs, kept inside because it is the common case.
```
jl-wynen: I thought it holds the graph from its inputs to outputs?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+Every held value is an object that the driver can inspect, clear, serialize, or place on a process boundary, which is the explicit lifetime that scipp/sciline#241 asks for.
+A value leaves one driver and enters another as a parameter of the flat pipeline: an aggregation over banks whose result feeds a `StreamProcessor` is `pipeline[EmptyDetector] = banks.compute(table)[EmptyDetector]` before the processor is built.
```
jl-wynen: 'driver' has not been defined. How can values move between drivers? I would understand a driver as a single piece of code that runs several stages.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
```diff
@@ -0,0 +1,523 @@
...
+agg.accumulators()              # fresh accumulators, one per accumulation key, for a caller that pushes
+agg.combine([...])              # push each contribution into fresh accumulators, read them
+agg.finalize(contribution)      # accumulation keys -> outputs
+agg.compute(table)              # contribute per row, pushing each as it is made; combine; finalize
+compute_members(pipeline, members=(Filename[SampleRun],), key=NormalizedQ[SampleRun, Numerator], table=table)
+```
+
+Semantics:
+
+- An aggregation is two stages of one flat pipeline with one accumulator per accumulation key between them: contribute from the member keys to the accumulation keys, and finalize from the accumulation keys to the outputs.
```
jl-wynen: How do `Aggregation`s compose if you have more than  stages?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
jl-wynen, line 1: I read about half of this document and frankly, the writing is terrible. The language is obtuse and convoluted and the content is poorly structured, mixing motivations, detailed descriptions of implementation, alternatives, etc. I feel like I need to have read the entire PR and some extra documents in several repos before I can understand it.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
SimonHeybrock: It referred to the table we `map` over and then reduce, now named `Aggregation` but `Fold` was the working class name and not cleaned up here. This term is avoided in the new version.

The three entry points meant the three functions on the `Aggregation` interface that allow running it "by hand", the `contribute` stage, the `combine` accumulation callable, the `finalize` stage. There is a drawing of this in the new version.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
SimonHeybrock: It does not belong here and is removed in the latest version. Q1 referred to a question/option in a design document that led to the two generics prototypes (both PRs closed), the "spike" was the quick experimental implementation of that:

> Q1: Forward chaining, or deferred mapped-labeling in cyclebane?

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
SimonHeybrock: I agree, it was still a working document, I apologize for not making that clear.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
SimonHeybrock: Currently I did not foresee composing aggregations. I someone needs a multi-level aggregation sth. should likely built for that purpose, similar to how `StreamProcessor` serves a specific purpose.

docs/developer/architecture-and-design/map-reduce-outside-the-graph.md
SimonHeybrock: It referred to the `combine` method/step of `class Aggregation`.

docs/developer/architecture-and-design/loki_validation.py
```diff
@@ -0,0 +1,353 @@
...
+SCHEDULER = sciline.scheduler.NaiveScheduler()
+
+calls: dict[str, int] = {}
+
+
+def _hit(name: str) -> None:
+    calls[name] = calls.get(name, 0) + 1
+
+
+# Counted copies of providers, with the original signatures so sciline sees them.
```
jl-wynen, line 72: You could also use https://scipp.github.io/sciline/generated/modules/sciline.reporter.TimingReporter.html#sciline.reporter.TimingReporter which returns the number of calls to each provider in `.as_pandas()`. (In case this code should be preserved.)

Except that we lose the ability to get a benchmark or progress report across stages. We would need to add that to each reduction class manually.

docs/developer/architecture-and-design/loki_validation.py
```diff
@@ -0,0 +1,353 @@
...
+        return Stage(self._pipeline, outputs=OUTPUTS, inputs=keys, scheduler=SCHEDULER)
+
+    def __setitem__(self, key: Any, value: Any) -> None:
+        self._pipeline[key] = value
+        for run_type, agg in self._aggregations.items():
+            if key in agg.contribute_stage.keys:
+                self._aggregations[run_type] = self._aggregation(run_type)
+                self._contributions[run_type].clear()
+        if key in self._finalize.keys:
+            self._finalize = self._finalize_stage()
```
jl-wynen: Why does this function set aggregations and the finalize stage?

docs/developer/architecture-and-design/loki_validation.py
```diff
@@ -0,0 +1,353 @@
...
+        for qtype in (NormalizedQ, NormalizedQxQy)
+    }
+
+
+def show(label: str, t0: float) -> dict[str, int]:
+    print(f'  {label}: {time.perf_counter() - t0:.1f} s, calls {calls}')
+    return dict(calls)
+
+
+class SansReduction:
```
jl-wynen, line 165: There is a bunch of non-trivial code in this class. Do we have to duplicate and adapt this for every workflow? Or can we have a common implementation?

jl-wynen, line 165: This looks too complicated for what is essentially a for-loop over runs if we inline mask handling in a single provider. Do you think this is worth the effort?

docs/developer/architecture-and-design/loki_validation.py
```diff
@@ -0,0 +1,353 @@
...
+    for key in OUTPUTS:
+        compare(key.__name__, staged[key], results[key])
+
+    print(f'\nSUMMARY: reference {t_ref:.1f} s, prototype {t_proto:.1f} s')
+    print(f'  reference calls: {ref_calls}')
+    print(f'  prototype calls: {proto_calls}')
+
+
+if __name__ == '__main__':
+    main()
```
nvaytet, line 342: I wasn't sure where to comment, so I put it here (so we can have replies attached to the question).

For Bifrost, we use map/reduce to map over the 45 detectors: https://github.com/scipp/ess/blob/main/packages/essspectroscopy/src/ess/bifrost/workflow.py#L159

How would this look/work with the new mechanism?

docs/developer/architecture-and-design/loki_validation.py
SimonHeybrock: I had intended to remove this before merging, this was just to validate we can do what we need downstream.

docs/developer/architecture-and-design/loki_validation.py
SimonHeybrock: I would hope we can make something reusable, but probably not everything fits into the same shape exactly.

SimonHeybrock: What I had in mind was to give an object with a "simple" interface to users (in a notebook) so they would not need to perform more error-prone coding in the notebook. I agree it would not be worth it for developers, but if we think about notebooks or other user interface...?

docs/developer/architecture-and-design/loki_validation.py
SimonHeybrock: I have not thought about it down to every detail, but map/reduce should essentially be replaced by `Aggregation`. The non-trivial bit is probably that the workflow might need some extra merged info and has some duplicate branches (see my current problems in scipp/esslivedata#1318).

docs/developer/architecture-and-design/loki_validation.py
SimonHeybrock: This was an attempt to make it behave like the old workflow (where map/reduce returned a new `Pipeline`). I actually think we should *not* do this as it is complicated and a bit confusing. What I would actually vote for is:
```
base_workflow = sciline.Pipeline(...)
base_workflow[Param] = param
workflow = SansReduction(base_workflow, ...)
workflow.set_runs(...)
# setting params on `workflow` is not allowed
# setting params on `base_workflow` has no effect on `workflow`
result = workflow.compute()
```

docs/developer/architecture-and-design/loki_validation.py
jl-wynen: That is only good for users as long as everything works out of the box. If it takes an expert to implement even a small change, then I don't think we gain anything.

docs/developer/architecture-and-design/loki_validation.py
jl-wynen: I want to try and remove this completely for Bifrost. I think it should be possible to load all detectors in a single provider and then operate with a data array with an added dimension instead of representing this on the workflow level.

docs/developer/architecture-and-design/loki_validation.py
SimonHeybrock: Can you clarify what you have in mind?

jl-wynen: Something along the lines of this:
```python
results = []
for filename in filenames:
  wf[Filename[SampleRun]] = filename
  results.append(wf.compute(IofQ))
i_of_q = sum(results)
```
It would re-run common providers but we could precompute them and store the results in the workflow if performance is an issue.

Overall, it would cof course be more complicated than those 5 lines. But I don't think we need to go as far as what you proposed here for such a simple use case.

SimonHeybrock: Sure, this change will not prevent using loops. But in practice you need two accumulators for SANS, and then call a "finalize" workflow with the results of the accumulators and some other parameters.

